PR #3, opened 2022-03-30
survey filter completed

Rick-Feng-u:


Opeyem1a:
The code architecture has progressed to the point where this doesn't serve a purpose anymore, sorry about that

PR #42, opened 2023-09-04
Minor enhancement for accessing id

ketphan02:
Created `get_id()` method to access protected field `_id` and update the usage


Opeyem1a:
I know why you made this PR, but I honestly don't think it matters too much, the only reason it's named `_id` is to avoid shadowing the outer `id` from python and we could just as easily solve this problem by renaming the variable.

If you think it's necessary I'll hear you out though - what are your thoughts?

Opeyem1a (approved):

PR #47, opened 2023-09-06
shuffle students returned from the mock student provider

Opeyem1a:


benchmarking/data/simulated_data/mock_student_provider.py
```diff
@@ -18,14 +18,19 @@ def __init__(self, settings: MockStudentProviderSettings):
...
             self.settings.number_of_friends,
             self.settings.number_of_enemies,
             self.settings.friend_distribution,
             self.settings.attribute_ranges,
             self.settings.num_values_per_attribute,
         )
+        # the students must be shuffled here because certain algorithms
+        #   perform better/worse based on the ordering of students.
+        #   i.e. if each sequential block of 5 students all selected one another as friends
+        random.shuffle(mock_students)
```
ketphan02, line 32: Shouldn't this shuffle be inside of `create_mock_students` ?

Opeyem1a: I want to leave each component as separate as possible from the others, so creating mock students should just be responsible for making them.

Shuffling is only necessary for statistical relevance -> i.e. only when the students being returned are explicitly going to be used for some sort of experiment. So I thought it'd be best to only shuffle them as a result of calling `.get()`

ketphan02: Ah got chu

ketphan02 (approved):

PR #48, opened 2023-09-06
Added new variable for mocking or not option

ketphan02:
### Issue
- The `test_saved_run()` method only runs with mock data
### Changes
- Added a new variable `is_mock` to determined the environment of the function
- Separated creating mock classes to functions
- **Minor changes:**
  - Added `students` as a property of `MockStudentProvider` and start creating list of students when initializing the mock
  - Modified the class accordingly
### Primary reviewer
@Opeyem1a

### DIFF
--- a/benchmarking/data/simulated_data/mock_student_provider.py
+++ b/benchmarking/data/simulated_data/mock_student_provider.py
@@ -16,9 +16,7 @@ from benchmarking.data.interfaces import (
 class MockStudentProvider(StudentProvider):
     def __init__(self, settings: MockStudentProviderSettings):
         self.settings = settings
-
-    def get(self) -> List[Student]:
-        return create_mock_students(
+        self.students: List[Student] = create_mock_students(
             self.settings.number_of_students,
             self.settings.number_of_friends,
             self.settings.number_of_enemies,
@@ -27,9 +25,12 @@ class MockStudentProvider(StudentProvider):
             self.settings.num_values_per_attribute,
         )
 
+    def get(self) -> List[Student]:
+        return self.students
+
     @property
     def num_students(self) -> int:
-        return self.settings.number_of_students
+        return len(self.students)
 
     @property
     def max_project_preferences_per_student(self) -> int:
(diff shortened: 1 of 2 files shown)


benchmarking/data/simulated_data/mock_student_provider.py
```diff
@@ -16,9 +16,7 @@
 class MockStudentProvider(StudentProvider):
     def __init__(self, settings: MockStudentProviderSettings):
         self.settings = settings
-
-    def get(self) -> List[Student]:
-        return create_mock_students(
+        self.students: List[Student] = create_mock_students(
```
Opeyem1a, line 19: This is pretty cool - I think we should leave as is though

It's important that the `.get()` returns a new set of students each time because the students might not be the same and this is on purpose. Various aspects of how we generate students is randomized and to preserve the statistical relevance of benchmarks we want to run multiple trials knowing that the set of students is slightly different.

benchmarking/runs/test_saved_run.py
```diff
@@ -75,3 +64,30 @@ def test_saved_run():
                 simulation_outputs, "NumTeamsMeetingRequirements"
             ),
         )
+
+
+def _get_student_provider_settings(is_mock: bool):
```
Opeyem1a, line 69: We can set this up as a convention, but remember that as we go, you and the team will be creating additional files within the `/runs` directory. So it might be better to just have separate files for mock vs real data runs as opposed to a single run doing both.

Also consider that with real data, the set of students is always the same in each trial whereas the set of students changes slightly per-trial with mock data - we might not want to lump these implications into a boolean flag like `is_mock`

Opeyem1a: With that said I like this approach in general so let's revisit the idea when we write our first few "real" files in the `/runs` directory

Opeyem1a (changes requested):

benchmarking/runs/test_saved_run.py
ketphan02: Should we create different folders for real/mock data then?

Opeyem1a: I believe there already are, right? /data/simulated_data vs /data/real_data

ketphan02:
Not necessary PR, should be closed

PR #50, opened 2023-09-07
Add Python Version to Readme and Main to Run test_saved_run

Hedgemon4:
## Changes
- Added python version to readme
- Added main function to test_saved_run
- Added requirement to convention that test files must have a main function

### DIFF
--- a/benchmarking/runs/CONVENTION.md
+++ b/benchmarking/runs/CONVENTION.md
@@ -2,4 +2,6 @@
 This directory contains simulation runs.
 Try to be descriptive when naming new files here.
 
+- Every file should contain a main function
+
 todo: complete
--- a/benchmarking/runs/CONVENTION.md
+++ b/benchmarking/runs/CONVENTION.md
@@ -2,6 +2,10 @@
 This directory contains simulation runs.
 Try to be descriptive when naming new files here.
 
-- Every file should contain a main function
+- Every file should have a "main clause" that allows them to be run from the terminal:
+```python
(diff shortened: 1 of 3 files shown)


benchmarking/runs/CONVENTION.md
```diff
@@ -2,4 +2,6 @@
 This directory contains simulation runs.
 Try to be descriptive when naming new files here.
 
+- Every file should contain a main function
```
Opeyem1a, line 5: Every file should have a "main clause" that allows them to be run from the terminal:
```python
if __name__ == "__main__":
    func()
```


My suggestion ^

Opeyem1a (approved):

benchmarking/runs/CONVENTION.md
ketphan02: I see a few projects use this pattern, not so different from Opey's suggestion
```python
def main():
    func()
if __name__ == "__main__": 
    main()
```

Opeyem1a: I kinda like the functions having names though, in case we make like "batch run" files that call many of them in sequence

Opeyem1a: Oh ohhh I see what you're saying now my bad

Opeyem1a: Idk if this is needed either but I kinda see why people do this? Let's leave as is for now

Opeyem1a (approved):

PR #53, opened 2023-09-08
Diversify social with friends

ketphan02:
### Issue
We want to create new scenario: diversify to satisfy every group has at least 2 friends

### 
- Added new metric `AverageSocialSatisfaction`
- Added new `TeamOption` to get the function for the new metric
- Added new Scenario
- Added new run test to validate

### Sample output
```text
CLASS SIZE / 100
=> {<AlgorithmType.RANDOM: 'random'>: 1.0, <AlgorithmType.WEIGHT: 'weight'>: 1.0, <AlgorithmType.SOCIAL: 'social'>: 1.0, <AlgorithmType.PRIORITY: 'priority'>: 1.0}
CLASS SIZE / 150
=> {<AlgorithmType.RANDOM: 'random'>: 1.0, <AlgorithmType.WEIGHT: 'weight'>: 1.0, <AlgorithmType.SOCIAL: 'social'>: 1.0, <AlgorithmType.PRIORITY: 'priority'>: 1.0}
CLASS SIZE / 200
=> {<AlgorithmType.RANDOM: 'random'>: 1.0, <AlgorithmType.WEIGHT: 'weight'>: 1.0, <AlgorithmType.SOCIAL: 'social'>: 1.0, <AlgorithmType.PRIORITY: 'priority'>: 1.0}
CLASS SIZE / 250
=> {<AlgorithmType.RANDOM: 'random'>: 1.0, <AlgorithmType.WEIGHT: 'weight'>: 1.0, <AlgorithmType.SOCIAL: 'social'>: 1.0, <AlgorithmType.PRIORITY: 'priority'>: 1.0}
CLASS SIZE / 300
=> {<AlgorithmType.RANDOM: 'random'>: 1.0, <AlgorithmType.WEIGHT: 'weight'>: 1.0, <AlgorithmType.SOCIAL: 'social'>: 1.0, <AlgorithmType.PRIORITY: 'priority'>: 1.0}
```


benchmarking/runs/run_diversify_2_friends.py
```diff
@@ -0,0 +1,81 @@
...
+    NumTeamsMeetingRequirements,
+)
+from benchmarking.evaluations.scenarios.diversify_social_min_2_friends import (
+    DiversifySocialMin2Friends,
+)
+from benchmarking.simulation.simulation import Simulation
+from models.enums import ScenarioAttribute, Relationship
+
+
+def copy_of_test_saved_run():
```
ketphan02, line 19: Change the name

models/enums.py
```diff
@@ -52,6 +52,7 @@ class ScenarioAttribute(Enum):
     YEAR_LEVEL = 6
     TIMESLOT_AVAILABILITY = 7
     PROJECT_PREFERENCES = 8
+    SOCIAL = 9
```
Opeyem1a, line 55: I'm not which attribute this is supposed to denote.

In general, ScenaroAttribute is basically just a way for us to give names to attribute ids we have "reserved" for these meanings. Think of each one as being something "about" a student.

So students have genders, year levels, project preferences, etc. This list should only include scenario attributes we aim to run experiments on.

Note: If you meant this to mean the student's social preferences, we actually specify that in a different way (not an attribute, check models/student.py and note the relationships field). Also you would need to use the PreferenceGoal with the "social" setting instead

benchmarking/runs/run_diversify_2_friends.py
```diff
@@ -0,0 +1,85 @@
...
+        )
+        print(
+            "=>",
+            Simulation.average_metric(
+                simulation_outputs, "NumTeamsMeetingRequirements"
+            ),
+        )
+
+
+def main():
```
Opeyem1a, line 80: nit;

I don't know what doing this affords us in the future, but it doesn't conflict with anything right now so if you want to do this, it works for me

benchmarking/runs/run_diversify_2_friends.py
```diff
@@ -0,0 +1,85 @@
...
+from benchmarking.evaluations.scenarios.diversify_social_min_2_friends import (
+    DiversifySocialMin2Friends,
+)
+from benchmarking.simulation.simulation import Simulation
+from models.enums import ScenarioAttribute, Relationship
+
+
+def run_diversify_2_friends():
+    """
+    Goal: Run diversify gender scenario, measure average gini index
```
Opeyem1a, line 21: These comments should be updated

nit; function name could be more descriptive too

benchmarking/runs/run_diversify_2_friends.py
```diff
@@ -0,0 +1,85 @@
...
+        number_of_teams = math.ceil(class_size / 5)
+
+        # set up either mock or real data
+        student_provider_settings = MockStudentProviderSettings(
+            number_of_students=class_size,
+            number_of_friends=4,
+            number_of_enemies=1,
+            friend_distribution="cluster",
+            attribute_ranges={
+                ScenarioAttribute.SOCIAL.value: [
```
Opeyem1a, line 42: Read later comment but tldr; this isn't how we specify friends/enemies, the `number_of_friends`/`number_of_enemies` is how we do that.

Also, every non-friend, non-enemy relationship is defaulted to be neutral in the algorithms themselves, so we don't need to mock that

Opeyem1a (commented):
I think you understand how to use the MockStudentProvider pretty well but I'm not sure I understand the scenario you picked and if I do, I don't think this is the way to run that experiment

Opeyem1a (changes requested):
Whoops forgot to request changes on the review

benchmarking/evaluations/metrics/utils/team_option.py
```diff
@@ -0,0 +1,43 @@
+from benchmarking.evaluations.metrics.utils.team_calculations import *
+
+
+class TeamOption:
```
Opeyem1a, line 4: nit;

This is super cool but feels kinda overkill imo? Like why not just pass a reference to the function you want to the AverageSocialSatisfaction directly?

i.e.
```python
AverageSocialSatisfaction(social_function=is_strictly_happy_team_enemy)
```

benchmarking/evaluations/scenarios/diversify_social_only.py
```diff
@@ -0,0 +1,21 @@
...
+
+from benchmarking.evaluations.enums import PreferenceDirection, PreferenceSubject
+from benchmarking.evaluations.interfaces import (
+    Scenario,
+    Goal,
+)
+from benchmarking.evaluations.goals import WeightGoal, PreferenceGoal
+
+
+class DiversifySocialFriends(Scenario):
```
Opeyem1a, line 11: I'm not sure the name of this describes the scenario still.

From the goals you've defined, it seems like this is `IncludeSocialFriends`

And the `def name()` should be more like "Create teams that group social friends together"
(the name should be descriptive)

benchmarking/runs/run_diversify_social_only.py
```diff
@@ -0,0 +1,90 @@
...
+)
+from benchmarking.evaluations.metrics.utils.team_option import TeamOption
+from benchmarking.evaluations.scenarios.diversify_social_only import (
+    DiversifySocialFriends,
+)
+from benchmarking.simulation.simulation import Simulation
+from models.enums import ScenarioAttribute, Gender
+
+
+def run_diversify_2_friends():
```
Opeyem1a, line 22: Same comments as with naming in the scenario applies to this function and file's name

I don't think the word "diversify" makes sense here

benchmarking/runs/run_diversify_social_only.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+    Goal: Run diversify gender scenario, measure average gini index
```
Opeyem1a, line 24: Update

benchmarking/runs/run_diversify_social_only.py
```diff
@@ -0,0 +1,90 @@
...
+                NumRequirementsSatisfied(),
+                NumTeamsMeetingRequirements(),
+            ],
+        ).run(num_runs=num_trials)
+
+        print(
+            "=>",
+            Simulation.average_metric(simulation_outputs, "AverageSocialSatisfaction"),
+        )
+        print(
```
Opeyem1a, line 77: These two print statements are not a part of this run, so they can be removed

benchmarking/runs/run_diversify_social_only.py
```diff
@@ -0,0 +1,90 @@
...
+            metrics=[
+                AverageSocialSatisfaction(
+                    TeamOption(
+                        strict_mode=False,
+                        friend_mode=True,
+                        is_all_happy=True,
+                        is_strictly_happy=False,
+                    )
+                ),
+                NumRequirementsSatisfied(),
```
Opeyem1a, line 68: No need to calculate metrics not relevant to the experiment.

I only added them in the test file to demonstrate the way to pass in multiple metrics

benchmarking/runs/run_diversify_social_only.py
```diff
@@ -0,0 +1,90 @@
...
+            number_of_students=class_size,
+            number_of_friends=4,
+            number_of_enemies=1,
+            friend_distribution="cluster",
+            attribute_ranges={
+                ScenarioAttribute.GENDER.value: [
+                    (Gender.MALE, 1 - ratio_of_female_students),
+                    (Gender.FEMALE, ratio_of_female_students),
+                ],
+                ScenarioAttribute.PROJECT_PREFERENCES.value: [1, 2, 3],
```
Opeyem1a, line 48: This part of the student provider settings isn't relevant to your experiment here, `attribute_ranges` and the `num_values_per_attribute` argument can be removed since your experiment doesn't need the students to be setup any further than having friends and enemies

Opeyem1a (changes requested):

benchmarking/evaluations/metrics/utils/team_option.py
ketphan02: Why couldn't I think of this 😭

Opeyem1a (approved):
Lgtm ✅

(we can discuss minor things like naming runs next meeting with the gang)

PR #58, opened 2023-09-12
add concentrate on multiple attributes scenario

Opeyem1a:
An idea to avoid calling it "all attributes"

I think it's important that scenario names are more accurate than it is for "run" names to be accurate, but it is to be discussed how to proceed with naming in this case.

i.e. if the intent of this run is always to concentrate on all attributes then it becomes the responsibility of whoever adds a new attribute to append it to this experiment.

alternatively

i.e. if we leave the scenario as hardcoding the goals then if you add a new attribute, it is your responsibility to then fix the "concentrate all attributes" scenario.

Discussion is around whether the scenarios are responsible for their accuracy or the runs

`Note for Opey:` Maybe we can rename "all attributes" to "many attributes", the reasoning being that the goal of the "experiment" is to see how they behave with many attributes. If we just keep adding attributes, we might not gain any new information (and if we wanted to see the effects of new attributes on each algorithm, that's a new experiment where the x-axis is number of attributes)


Opeyem1a:
* main
  * **PR #58** <a href="https://app.graphite.dev/github/pr/Teamable-Analytics/algorithms/58" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/Teamable-Analytics/algorithms/58?utm_source=stack-comment).

justino599:
I like this better as it provides expandable usability, and the name is true to what it does as we add more attributes. This way the run only needs to be updated if there is a need for it to be updated, not every time a new attribute is added. It's probably not a good idea to have parts of the code that need to be updated manually if you decide to add something elsewhere as that would quickly become unmanageable if we were to add more dependencies like that.

tldr; I'm in favour of merging this

PR #70, opened 2023-09-19
Add Graphs to Concentrate GPA Scenario

Hedgemon4:
# Changes
Added graphs to concentrate GPA scenario.

# Graphs
image
image
image
image


benchmarking/runs/concentrate_gpa_run.py
```diff
@@ -61,6 +76,85 @@ def concentrate_gpa_run():
             "Minimum Gini Index for GPA =>",
             Simulation.average_metric(simulation_outputs, "MinimumGiniIndex"),
         )
+        print(
+            "Runtimes =>",
+            Simulation.average_metric(simulation_outputs, "runtimes"),
```
Opeyem1a, line 81: nit; 

technically better to use `Simulation.KEY_RUNTIMES` instead of the string

benchmarking/runs/concentrate_gpa_run.py
```diff
@@ -61,6 +76,85 @@ def concentrate_gpa_run():
             "Minimum Gini Index for GPA =>",
             Simulation.average_metric(simulation_outputs, "MinimumGiniIndex"),
         )
+        print(
```
Opeyem1a, line 79: Tbh I say just kill the prints if we have graphs

benchmarking/runs/concentrate_gpa_run.py
```diff
@@ -61,6 +76,85 @@ def concentrate_gpa_run():
...
+            x_lim=None,
+        )
+    )
+
+    line_graph(
+        LineGraphMetadata(
+            x_label="Class size",
+            y_label="Maximum Gini Index",
+            title="Concentrate GPA Max Gini",
+            data=list(graph_dicts[3].values()),
```
Opeyem1a, line 152: the dictionaries are named variables, why access them from the list index?

Opeyem1a (changes requested):

Opeyem1a (approved):

PR #72, opened 2023-09-20
Generate graphs of runtime and avg gini index for many attributes scenario run

justino599:
### Goal
- Generate graph for currently printed statistics: Average Gini index
- Generate graph for average simulation runtime

### Result
**Note:** The runtimes are a little bit skewed here because I ran it on my laptop while running around campus

image
image


benchmarking/runs/concentrate_many_attributes.py
```diff
@@ -16,14 +19,21 @@
 
 def concentrate_many_attributes():
     """
-    Goal: Run concentrate on many attributes scenario (6 attributes), measure average gini index across all attributes
+    Goal: Run concentrate on many attributes scenario (6 attributes), measure average gini index across many attributes
```
Opeyem1a, line 22: Nice catch

Opeyem1a (approved):

PR #80, opened 2023-09-24
run algorithms in parallel

### LINKED ISSUE #81: Parallelize algorithm runs

Opeyem1a:
Goals:
- uses the `threading` library to run each algorithm in parallel (per trial)
- allows us to up some of the default parameters on the priority algorithm (so it has more of a chance to actually do things)

Validated:
- Ran a run in parallel and in sequence, as expected, the sequential simulation took twice as long
Rationale: priority and priority_new take the longest (by far) to run and we run both so parallel saved us the exact amount of time that the extra generation from priority was causing

closes #81

### DIFF
--- a/benchmarking/simulation/simulation.py
+++ b/benchmarking/simulation/simulation.py
@@ -1,5 +1,6 @@
 import copy
 import statistics
+import threading
 import time
 from collections import defaultdict
 from typing import List, Dict, Union
@@ -12,7 +13,11 @@ from benchmarking.data.interfaces import (
 )
 from benchmarking.evaluations.interfaces import Scenario, TeamSetMetric
 from benchmarking.simulation.algorithm_translator import AlgorithmTranslator
+from benchmarking.simulation.mock_algorithm import MockAlgorithm
+from api.models.enums import AlgorithmType
 from old.team_formation.app.team_generator.algorithm.algorithms import AlgorithmOptions
+from old.team_formation.app.team_generator.student import Student as AlgorithmStudent
+from old.team_formation.app.team_generator.team_generator import TeamGenerationOption
 
 RunOutput = Dict[AlgorithmType, Dict[str, List[float]]]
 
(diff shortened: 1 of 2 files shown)


Opeyem1a:
* main
  * **PR #80** <a href="https://app.graphite.dev/github/pr/Teamable-Analytics/algorithms/80" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/Teamable-Analytics/algorithms/80?utm_source=stack-comment).

ai/priority_algorithm/priority_algorithm.py
```diff
@@ -20,8 +20,8 @@ class PriorityAlgorithm(WeightAlgorithm):
 
     MAX_KEEP: int = 3  # nodes
     MAX_SPREAD: int = 3  # nodes
-    MAX_ITERATE: int = 15  # times
-    MAX_TIME: int = 30  # seconds
+    MAX_ITERATE: int = 1500  # times
+    MAX_TIME: int = 7  # seconds
```
Hedgemon4, line 24: Are these just arbitrary numbers or is there a reason you choose these?

Hedgemon4 (changes requested):

ai/priority_algorithm/priority_algorithm.py
Opeyem1a: Arbitrary, might have gotten a bit ambitious given the local dev experience

Opeyem1a:
@Hedgemon4 Are these numbers doable on local for you? If not we can always just set the numbers back to normal and have this PR actually be *just* parallelizing them haha

justino599:
It seems runnable, but it is definitely not utilizing my system any better than main. My system has a 4 core/8 thread cpu.

On this branch:
image
On main:
image

Opeyem1a:
@justino599 Suggested parallelizing on the trial level instead of the algorithm level, I'll do this instead

EliasPinno:
@Opeyem1a and I discussed this ticket. Here is a concept for a solution.
```

for x in x_var -> for example, class in class sizes
    before you fork, see if an in progress run existed for this x.
    if it did, join all the in progress results, and count how many jsons are in the final file.
    this will tell you how high trials should be for this run.
    fork here: split amongst number of threads
    each thread opens their own file to write to
    for n in trials -> trials is a number, n is just a counter
         for algo_type in types:
             gen_using_algo. Then write that json to this threads file
     join here
     combine the 4 files into 1 file with all the json generated results. Append all the jsons to the existing file (we create it before fork?)
```

PR #85, opened 2023-09-26
Create Robinhood mutation method for priority algorithm

### LINKED ISSUE #76: Implement "Robin Hood" mutate method for priority algorithm
### Issue
- Mutations are random right now and should be done in a smarter way

### Suggested Fix
- Implement the "Robin Hood" mutation method:
	- Choose arbitrary* constraint c and find a team that does not satisfy c. Find another team that does satisfy c as the two teams. Then do the brute force step in local max.
	- *Repeat this for each constraint and generate a teamset resulting from doing this on each constraint.
If there are 3 constraints, do this exchange 3 times, and the algorithm would get 3 resulting teamsets (keep 3 nodes in the expanding tree).
	- More details: https://docs.google.com/document/d/1e6fVKQmt6a9qyHZNVqCP5TCugaAKtr8RzV9AjOmimYo/edit#heading=h.psoadlxedk50

justino599:
### Goal
- Implement the "Robinhood" mutation method
	- Swaps students between a team that satisfies a constraint and a team that does not
	- See #76 for more

Closes #76

### DIFF
--- a/ai/priority_algorithm/mutations.py
+++ b/ai/priority_algorithm/mutations.py
@@ -144,6 +144,9 @@ def mutate_robinhood(
                 ]
 
                 # Calculate the score of the resulting PriorityTeamSet
+                cloned_priority_team_set.score = (
+                    None  # Reset the score to force recalculation after modification
+                )
                 score = cloned_priority_team_set.calculate_score(
                     priorities, student_dict
                 )
--- a/ai/priority_algorithm/mutations.py
+++ b/ai/priority_algorithm/mutations.py
@@ -49,6 +49,41 @@ def mutate_robinhood(
     Robinhood is a mutation that finds a team t1 that does not satisfy a priority c, and a team t2 that does satisfy c.
     It then creates all possible teams using the students of t1 and t2, and chooses the M best teams.
     """
+
+    # Argument checking
+    if len(priority_team_set.priority_teams) == 0:
+        raise ValueError("PriorityTeamSet must have at least one team")
+    if len(priority_team_set.priority_teams) == 1:
+        return priority_team_set
+    if len(priorities) == 0:
+        raise ValueError("Must have at least one priority")
+    all_students: List[Student] = [
+        student
+        for team in priority_team_set.priority_teams
+        for student in team.team.students
+    ]
+    if len(all_students) != len(student_dict):
+        raise ValueError("student_dict must contain all students in priority_team_set")
+    for student in all_students:
+        if student.id not in student_dict:
+            raise ValueError(
+                "student_dict must contain all students in priority_team_set"
+            )
+    if len(all_students) == 0:
+        raise ValueError("PriorityTeamSet must have at least one student")
+    if (
+        len(
+            [
+                team
+                for team in priority_team_set.priority_teams
+                if not team.team.is_locked
+            ]
+        )
+        < 2
+    ):
+        return priority_team_set
+
+    # Init best team set
     best_team_set: PriorityTeamSet = priority_team_set
     best_team_set_score: float = priority_team_set.calculate_score(
         priorities, student_dict
@@ -76,8 +111,18 @@ def mutate_robinhood(
                     unsatisfied_teams.append(team)
 
             # Choose a random team from each list
-            satisfied_team: PriorityTeam = random.choice(satisfied_teams)
-            unsatisfied_team: PriorityTeam = random.choice(unsatisfied_teams)
+            if len(unsatisfied_teams) == 0 or len(satisfied_teams) == 0:
+                satisfied_team = random.choice(available_priority_teams)
+                unsatisfied_team = random.choice(
+                    [
+                        team
+                        for team in available_priority_teams
+                        if team != satisfied_team
+                    ]
+                )
+            else:
+                satisfied_team = random.choice(satisfied_teams)
+                unsatisfied_team = random.choice(unsatisfied_teams)
 
             # List of all students in the two teams
             students: List[int] = (
--- a/ai/priority_algorithm/mutations.py
+++ b/ai/priority_algorithm/mutations.py
@@ -46,8 +46,7 @@ def mutate_robinhood(
     student_dict: Dict[int, Student],
 ) -> PriorityTeamSet:
     """
-    Robinhood is a mutation that finds a team t1 that does not satisfy a priority c, and a team t2 that does satisfy c.
-    It then creates all possible teams using the students of t1 and t2, and chooses the M best teams.
+    Robinhood is a mutation that finds a team t1 that does not satisfy a priority c, and a team t2 that does satisfy c. It then creates all possible team sets by mutating the students of t1 and t2, and chooses the best team. This is done across all constraints.
     """
 
     # Argument checking
--- a/ai/priority_algorithm/mutations.py
+++ b/ai/priority_algorithm/mutations.py
@@ -50,36 +50,8 @@ def mutate_robinhood(
     """
 
     # Argument checking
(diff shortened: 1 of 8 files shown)


ai/priority_algorithm/mutations.py
```diff
@@ -30,3 +38,186 @@ def mutate_random_swap(priority_team_set: PriorityTeamSet) -> PriorityTeamSet:
...
+        score = cloned_priority_team_set.calculate_score(priorities, student_dict)
+        if score > best_team_set_score:
+            # Clone the PriorityTeamSet because we will be modifying on the next iteration
+            best_team_set = cloned_priority_team_set.clone()
+            best_team_set_score = score
+
+    return best_team_set, best_team_set_score
+
+
+def valid_robinhood_arguments(
```
Opeyem1a, line 168: this is cool but most of these cases are either checked before it gets here or don't matter, I'll leave some more comments but tldr is if the mutation would do nothing then just return the unedited teamset (so don't "mutate" it at all)

ai/priority_algorithm/mutations.py
```diff
@@ -30,3 +38,186 @@ def mutate_random_swap(priority_team_set: PriorityTeamSet) -> PriorityTeamSet:
...
+    priority_team_set: PriorityTeamSet,
+    priorities: List[Priority],
+    student_dict: Dict[int, Student],
+) -> bool:
+    if len(priority_team_set.priority_teams) == 0:
+        raise ValueError("PriorityTeamSet must have at least one team")
```
Opeyem1a, line 174: Like in this case, just return the teamset

ai/priority_algorithm/mutations.py
```diff
@@ -30,3 +38,186 @@ def mutate_random_swap(priority_team_set: PriorityTeamSet) -> PriorityTeamSet:
...
+    if len(priority_team_set.priority_teams) == 1:
+        return False
+    if len(priorities) == 0:
```
Opeyem1a, line 177: If this was the case, just return the teamset and the priority algorithm will run fasst (doing nothing), but the API to the algorithms will enforce specifying a priority if using this algorithm

ai/priority_algorithm/priority_algorithm.py
```diff
@@ -148,9 +148,17 @@ def mutate(self, team_set: PriorityTeamSet) -> List[PriorityTeamSet]:
         """
         Mutate a single teamset into child teamsets
         """
+        algorithm = 1
```
Opeyem1a, line 151: note;
I'll clean the way we do this at some point in the future, the direction isn't aligned with what I recommended to @Hedgemon4 to solve this problem, whichever one of you merges first will have to bully the other to some temporary consensus

tests/test_ai/test_priority_algorithm/test_mutations.py
```diff
@@ -0,0 +1,263 @@
...
+
+    def validate(self):
+        return True
+
+    def satisfied_by(self, students: List[Student]) -> bool:
+        ids = [student.id for student in students]
+        return 1 in ids and 2 in ids
+
+
+def print_priority_team_set(priority_team_set: PriorityTeamSet):
```
Opeyem1a, line 25: probably remove now right?

tests/test_ai/test_priority_algorithm/test_mutations.py
```diff
@@ -0,0 +1,263 @@
...
+    """
+    For debugging purposes
+    """
+    for priority_team in priority_team_set.priority_teams:
+        print(f"Team {priority_team.team.id}: {priority_team.student_ids}", end=" ")
+    print()
+
+
+def equal_priority_team_sets(a: PriorityTeamSet, b: PriorityTeamSet) -> bool:
```
Opeyem1a, line 34: I don't think this method 100% does what you think, it seems to assume the student ids are ordered, if you want to make this better, I would look into `symmetric_difference` after turning both student id lists into sets

tests/test_ai/test_priority_algorithm/test_mutations.py
```diff
@@ -0,0 +1,263 @@
...
+        missing_students = 0
+        for i in range(9):
+            if i not in all_students:
+                missing_students += 1
+
+        self.assertEqual(
+            missing_students, 0, "Not all students are in the mutated team set"
+        )
+
+    def test_mutate_robinhood__handles_invalid_arguments(self):
```
Opeyem1a, line 210: note;
yikes my bad, looks like this test is juicy but I don't think mutations should really be doing any "validation" at all, thus no such thing as handling invalid arguments

tests/test_ai/test_priority_algorithm/test_mutations.py
```diff
@@ -0,0 +1,263 @@
...
+        mutated_team_set = mutate_robinhood(priority_team_set, priorities, student_dict)
+
+        # Count the number of students in the original team set that are not in the mutated team set
+        all_students = [
+            student.id
+            for priority_team in priority_team_set.priority_teams
+            for student in priority_team.team.students
+        ]
+
+        assert len(all_students) == 9
```
Opeyem1a, line 199: q;
why not `self.assertEquals()` here? Pure curiousity

tests/test_ai/test_priority_algorithm/test_mutations.py
```diff
@@ -0,0 +1,263 @@
...
+        )
+    if priority_team_set.priority_teams[team_id].team.id == team_id:
+        return priority_team_set.priority_teams[team_id]
+    for priority_team in priority_team_set.priority_teams:
+        if priority_team.team.id == team_id:
+            return priority_team
+    raise ValueError(f"Team {team_id} not found")
+
+
+class TestMutations(unittest.TestCase):
```
Opeyem1a, line 88: tbh I would give testing each mutation it's own class for testing if it's going to be this juicy

tests/test_ai/test_priority_algorithm/test_mutations.py
```diff
@@ -0,0 +1,263 @@
...
+        if priority_team_a.team.id != priority_team_b.team.id:
+            return False
+        if priority_team_a.student_ids != priority_team_b.student_ids:
+            return False
+    return True
+
+
+def create_new_priority_team_set(
+    num_teams: int, num_students: int
+) -> (PriorityTeamSet, Dict[int, Student]):
```
Opeyem1a, line 50: nit;
You can import `Tuple` from the `typing` library instead of just using parentheses (it's supposed to be better but I forgot why, either way, it's more explicit)

ai/priority_algorithm/mutations.py
```diff
@@ -30,3 +38,186 @@ def mutate_random_swap(priority_team_set: PriorityTeamSet) -> PriorityTeamSet:
...
+    # Calculate the score of each team
+    for team in possible_teams:
+        # Modify the cloned PriorityTeamSet to reflect the new team
+        selected_team_b.student_ids = team
+        selected_team_a.student_ids = [
+            student_id for student_id in students if student_id not in team
+        ]
+
+        # Calculate the score of the resulting PriorityTeamSet
+        cloned_priority_team_set.score = (
```
Opeyem1a, line 156: note;
alas the tears of cache invalidation

Opeyem1a (changes requested):
wrote a lot of comments, but shouldn't be crazy changes

tests/test_ai/test_priority_algorithm/test_mutations.py
justino599: Probably because I forgot to switch it over after I figured out how to use `unittest` haha

Opeyem1a (approved):
lgtm

PR #103, opened 2023-09-28
98 implement epsilon local max mutation

Hedgemon4:


Hedgemon4:
Implemented the best version in #191

PR #114, opened 2023-09-29
Added a value error for 0 length metrics.

EliasPinno:


benchmarking/simulation/simulation.py
```diff
@@ -54,6 +53,8 @@ def __init__(
             raise ValueError(
                 "Either num_teams OR a project initial_teams_provider must be specified."
             )
+        if len(self.metrics) < 1:
```
Opeyem1a, line 56: nit;
I think the _pythonic_ way to do this is more like
```suggestion
        if not self.metrics:
```
Then we handle the None case implicitly too :D

Opeyem1a (changes requested):

benchmarking/simulation/simulation.py
EliasPinno: Done.

Opeyem1a (approved):

PR #117, opened 2023-10-05
Add api structure

ketphan02:


Opeyem1a (approved):
I was present while this was being written - lgtm

PR #118, opened 2023-10-05
Implement Priority Satisfaction Metric

Hedgemon4:
### Description
- implement priority satisfaction metric


benchmarking/evaluations/metrics/priority_satisfaction.py
```diff
@@ -0,0 +1,52 @@
...
+            else self.computer_exponential_weights()
+        )
+        total_score = 0
+        for team in team_set.teams:
+            priorities = self.priorities_satisfied(team)
+            product = sum(w * p for w, p in zip(weights, priorities))
+            total_score += product
+        # Normalize the scores between 0 and 1 for easier comparison
+        if self.is_linear:
+            total_score /= team_set.num_teams * sum(range(1, len(self.priorities) + 1))
```
Opeyem1a, line 32: ngl I have no idea how this normalizes, might just be sleepy opey doing bad math but code seems slightly unintuitive to me

Maybe it would help to have a `theoretical_max` variable defined above and then this like is just `total_score / theoretical_max`

Opeyem1a (changes requested):

benchmarking/simulation/goal_to_priority.py
```diff
@@ -0,0 +1,42 @@
...
+        return TokenizationConstraintDirection.MAX_OF
+    raise TypeError
+
+
+def goals_to_priorities(goals: List[Goal]) -> List[Priority]:
+    return [goal_to_priority(goal) for goal in goals]
+
+
+def goal_to_priority(goal: Goal) -> Priority:
+    if not isinstance(goal, DiversityGoal):
```
Opeyem1a, line 32: Also check if the diversity goal has a tokenization constraint (it has to to become a priority rn)

Opeyem1a (approved):
lgtm

PR #126, opened 2023-10-06
migrate weight algorithm utilities to new module

### LINKED ISSUE #88: Migrate weight algorithm utility functions to /new
Since it's using different dataclasses for Student. Team, etc, these utility methods need to be slightly modified to work in the new weight algorithm

Opeyem1a:
closes #88

No tests because they are technically already written in the web-app and #90 will bring them in

### DIFF
--- a/api/ai/new/weight_algorithm/utility/__init__.py
+++ b/api/ai/new/weight_algorithm/utility/__init__.py
@@ -1,0 +1,4 @@
+from .diversity_utility import get_diversity_utility
+from .preference_utility import get_preference_utility
+from .requirement_utility import get_requirement_utility
+from .social_utility import get_social_utility
--- a/api/ai/new/weight_algorithm/utility/social_network.py
+++ b/api/ai/new/weight_algorithm/utility/social_network.py
@@ -1,0 +1,166 @@
+from typing import List, Dict
+
+from api.models.enums import Relationship, Weight
+from api.models.student import Student
+
+
+class SocialNetwork:
+    _network = {}
+
+    def __init__(self, students: List[Student], balanced=True):
+        """
+        WARNING: Passing an unbalanced social network to SubSocialNetwork causes inconsistencies in its methods.
+        The ability to do this is currently only used for testing purposes.
+        """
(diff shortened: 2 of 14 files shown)


Opeyem1a:
* main
  * **PR #126** <a href="https://app.graphite.dev/github/pr/Teamable-Analytics/algorithms/126" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/Teamable-Analytics/algorithms/126?utm_source=stack-comment).

api/ai/new/weight_algorithm/utility/__init__.py
justino599, line 1: Why are we adding stuff into an `__init__.py`? I guess my question is really, what does this do? Because I'm sure there's a reason

api/ai/new/weight_algorithm/utility/social_network.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+class SocialNetwork:
+    _network = {}
+
+    def __init__(self, students: List[Student], balanced=True):
+        """
+        WARNING: Passing an unbalanced social network to SubSocialNetwork causes inconsistencies in its methods.
+        The ability to do this is currently only used for testing purposes.
+        """
```
justino599, line 14: Is this still the case?

justino599 (approved):
LGTM

api/ai/new/weight_algorithm/utility/__init__.py
Opeyem1a: @justino599 
Oh it just makes imports nicer, but in a high level way, it truly makes this directory into a "module"

So it's `from utility import social_utility` instead of `from utility.social_utility import social_utility`
In the same way you do `from typing import List` and not `from typing.lists import List`

api/ai/new/weight_algorithm/utility/social_network.py
Opeyem1a: Oh good point, I'll ponder

Issue #108, opened 2023-09-28
Develop a way to configure PriorityAlgorithm with different configs


Opeyem1a:
"Configs" refers to the constants like `MAX_TIME` and `MAX_SPREAD` that are defined within the `PriorityAlgorithm` class. These constants affect the algorithm significantly and we want to be capable of making a graph like "priority satisfaction (y-axis) vs MAX_SPREAD (x-axis)"

Idea:
- Some kind of `algorithm_configs` dict as a parameter to `Simulation`
- Each algorithm takes an instance of `Config` in their `__init__` and this is extended to be specific to algorithms (like the new `AlgorithmOption` interface is
- If config is provided in the init, it is used instead of the default config that will still be a static attribute in the algorithm classes

Note:
- Only priority currently needs a config, but it's conceivable that other algorithms might need this extensibility eventually (i.e. config for social algorithm might include whether to deal with remaining students using random or weight algorithm, for example)

PR #128, opened 2023-10-06
early work for algorithm config support

Opeyem1a:
related to #108 

Adds this structure for the reimplementation of the algorithms
Comments in the code specify, but tldr Configs exist so we can augment how algorithms actually behave and they are not something to be customized by end users, but eventually there will be a simple way to overwrite these configs within a simulation for benchmarking purposes

### DIFF
--- a/api/ai/new/interfaces/algorithm.py
+++ b/api/ai/new/interfaces/algorithm.py
@@ -14,7 +14,7 @@ class Algorithm(ABC):
         self,
         algorithm_options: AlgorithmOptions,
         team_generation_options: TeamGenerationOptions,
-            algorithm_config: AlgorithmConfig = None,
+        algorithm_config: AlgorithmConfig = None,
     ):
         self.algorithm_options: AlgorithmOptions = algorithm_options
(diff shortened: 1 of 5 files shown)


Opeyem1a:
* main
  * **PR #128** <a href="https://app.graphite.dev/github/pr/Teamable-Analytics/algorithms/128" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/Teamable-Analytics/algorithms/128?utm_source=stack-comment).

ketphan02:
As I understand this, we will pass `AlgorithmOption` instead of *args and **kwargs

justino599 (approved):
As this is "early work", I'll approve it, but I assume all the empty config classes will be filled out with their default configurations?

Opeyem1a:
@justino599 yeah, although some may remain empty/be deleted since not all algorithms currently have/need configs - I hadn't chosen yet and thought it would be nice to keep PRs small and progress moving

PR #132, opened 2023-10-09
Improve graphing tool

### LINKED ISSUE #131: Improve graphing tool
### Issues
- Every line should be distinguishable by 3 characteristics: colour, style (solid line, dashed line, etc.), marker (*, o, x, ...)
- y-axis should be the same (e.g. 0 to 1, or 0 to 100%) 
- Add minor ticks to x-axis - allow user to customize xticks and y-axis if necessary

justino599:
### Issue
1. Every line should be distinguishable by 3 characteristics: colour, style (solid line, dashed line, etc.), marker (*, o, x, ...)
2. y-axis should be the same (e.g. 0 to 1, or 0 to 100%)
3. Add minor ticks to x-axis - allow user to customize xticks and y-axis if necessary

### Fix
- [x] Automatically vary line styles and markers
	- [x] Create list of different line styles
	- [x] Create list of different line markers
- [x] Standardize y-axis range
	- [x] Add suggested y-axis range to each `TeamSetMetric`
	- [x] Update runs to use suggested ranges
- [x] Add minor ticks to x-axis

### Screenshots
Line styles
image

Closes #131


justino599:
To keep this PR up-to-date with private discussions, Opey and I discussed issue 2 and came to the following conclusions:
- The programmer should still ultimately have the control of what the y-axis range should be (e.g. a range of 0 to 1 should not be enforced for gini index)
	- A motivating example for why we thought this: "let's say for some cool new metric *in only 1 scenario*, all the values for all algorithms are super low (1-2) but the theoretical max of that metric is high (100), maybe we want that run specifically to graph them from 1-3 just so we can see more closely"
- There should still be a suggested range that will be used most of the time
	- Add to the `TeamSetMetric` interface a "theoretical min/max" value for the metric that can be used by the programmer as the y-axis limits

benchmarking/runs/concentrate_gpa.py
```diff
@@ -23,7 +25,7 @@ def concentrate_gpa(num_trials: int = 10):
     """
 
     # Define changing values
-    class_sizes = [50, 100, 150, 200, 250, 300, 350, 400]
+    class_sizes = [50, 100, 150]
```
Opeyem1a, line 28: I'm assuming this was a temp change

benchmarking/evaluations/graphing/line_graph.py
```diff
@@ -56,6 +73,13 @@ def line_graph(graph_data: LineGraphMetadata):
     if graph_data.y_label is not None:
         plt.ylabel(graph_data.y_label)
 
+    if graph_data.num_minor_ticks:
+        ax[0][0].xaxis.set_minor_locator(
+            AutoMinorLocator(graph_data.num_minor_ticks + 1)
+        )
+    else:
+        ax[0][0].xaxis.set_minor_locator(AutoMinorLocator())
```
Opeyem1a, line 81: joke; you have a way to auto locate minors? 🤨

benchmarking/runs/concentrate_gpa.py
justino599: Ah, yes it was, good catch!

benchmarking/evaluations/graphing/line_graph.py
justino599: Haha 😂

Opeyem1a (approved):
lgtm

PR #133, opened 2023-10-10
Implement Scenario With Multiple Tokenization Priorities

Hedgemon4:
### Description

- Implemented a scenario with multiple tokenization priorities for benchmarking the algorithms with priority satisfaction

### DIFF
--- a/benchmarking/evaluations/scenarios/multiple_tokenization_goals.py
+++ b/benchmarking/evaluations/scenarios/multiple_tokenization_goals.py
@@ -1,0 +1,49 @@
+from typing import List
+
+from api.models.enums import DiversifyType, ScenarioAttribute, TokenizationConstraintDirection
+from api.models.tokenization_constraint import TokenizationConstraint
+from benchmarking.evaluations.goals import DiversityGoal, WeightGoal
+from benchmarking.evaluations.interfaces import Scenario, Goal
+
+
+class MultipleTokenizationGoals(Scenario):
+    def __init__(self, value_of_gpa: int, value_of_race: int, value_of_age):
+        self.value_of_gpa = value_of_gpa
+        self.value_of_race = value_of_race
+        self.value_of_age = value_of_age
+
+    def name(self) -> str:
+        return "Scenario to run multiple tokenization goals"
+
+    @property
+    def goals(self) -> List[Goal]:
+        return [
+            DiversityGoal(
+                DiversifyType.CONCENTRATE,
+                ScenarioAttribute.GPA.value,
+                tokenization_constraint=TokenizationConstraint(
+                    direction=TokenizationConstraintDirection.MAX_OF,
+                    threshold=3,
+                    value=self.value_of_gpa
+                )
+            ),
+            DiversityGoal(
+                DiversifyType.DIVERSIFY,
+                ScenarioAttribute.RACE.value,
+                tokenization_constraint=TokenizationConstraint(
+                    direction=TokenizationConstraintDirection.MIN_OF,
+                    threshold=2,
+                    value=self.value_of_race,
+                ),
+            ),
+            DiversityGoal(
+                DiversifyType.CONCENTRATE,
+                ScenarioAttribute.AGE.value,
+                tokenization_constraint=TokenizationConstraint(
+                    direction=TokenizationConstraintDirection.MAX_OF,
+                    threshold=3,
+                    value=self.value_of_age
+                )
+            ),
+            WeightGoal(diversity_goal_weight=1)
+        ]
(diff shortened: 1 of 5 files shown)


benchmarking/evaluations/scenarios/multiple_tokenization_goals.py
```diff
@@ -0,0 +1,53 @@
...
+                    threshold=3,
+                    value=self.value_of_gpa,
+                ),
+            ),
+            DiversityGoal(
+                DiversifyType.DIVERSIFY,
+                ScenarioAttribute.RACE.value,
+                tokenization_constraint=TokenizationConstraint(
+                    direction=TokenizationConstraintDirection.MIN_OF,
+                    threshold=2,
```
Opeyem1a, line 39: why is this threshold different?

benchmarking/runs/multiple_tokenization_goals.py
```diff
@@ -0,0 +1,188 @@
...
+from benchmarking.evaluations.metrics.minimum_gini_index import MinimumGiniIndex
+from benchmarking.evaluations.metrics.priority_satisfaction import PrioritySatisfaction
+from benchmarking.evaluations.scenarios.multiple_tokenization_goals import (
+    MultipleTokenizationGoals,
+)
+from benchmarking.simulation.goal_to_priority import goals_to_priorities
+from benchmarking.simulation.simulation import Simulation
+
+
+def concentrate_multiple_max_3(num_trials: int = 10):
```
Opeyem1a, line 28: this name feels weird? I think don't be afraid of long function names here and just say the tokenization constraint part

benchmarking/runs/multiple_tokenization_goals.py
```diff
@@ -0,0 +1,188 @@
...
+    """
+    Goal: Run a scenario with multiple tokenization constraints to benchmark the algorithm
+    """
+
+    # Defining our changing x-values (in the graph sense)
+    class_sizes = list(range(50, 501, 50))
+    num_trials = 10
```
Opeyem1a, line 35: unneeded, num_trials comes from the function arguments
(pls quick check other runs in case we made this mistake elsewhere too)

benchmarking/runs/multiple_tokenization_goals.py
```diff
@@ -0,0 +1,188 @@
...
+from benchmarking.evaluations.scenarios.multiple_tokenization_goals import (
+    MultipleTokenizationGoals,
+)
+from benchmarking.simulation.goal_to_priority import goals_to_priorities
+from benchmarking.simulation.simulation import Simulation
+
+
+def concentrate_multiple_max_3(num_trials: int = 10):
+    """
+    Goal: Run a scenario with multiple tokenization constraints to benchmark the algorithm
```
Opeyem1a, line 30: "to benchmark ..." is not relevant, that's assumed, what might be relevant is mentioning the metrics you're using

benchmarking/runs/multiple_tokenization_goals.py
```diff
@@ -0,0 +1,188 @@
...
+    graph_dicts = [
+        graph_runtime_dict,
+        graph_avg_gini_dict,
+        graph_priority_dict,
+    ]
+
+    for class_size in class_sizes:
+        print("CLASS SIZE /", class_size)
+
+        number_of_teams = math.ceil(class_size / 5)
```
Opeyem1a, line 65: The team size here actually matters quite a bit, just me noting that you might want to be a bit more cognizant of this line making the team_size=5

benchmarking/evaluations/scenarios/multiple_tokenization_goals.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class MultipleTokenizationGoals(Scenario):
+    def __init__(self, value_of_gpa: int, value_of_race: int, value_of_age):
+        self.value_of_gpa = value_of_gpa
+        self.value_of_race = value_of_race
+        self.value_of_age = value_of_age
+
+    def name(self) -> str:
+        return "Scenario to run multiple tokenization goals"
```
Opeyem1a, line 20: you can be more specific here, intent of this name was to be able to use it while generating figures

Opeyem1a (changes requested):

benchmarking/evaluations/scenarios/multiple_tokenization_goals.py
Hedgemon4: I think this was just because I basically just wanted a scenario with more varied parameters to see how the algorithm performed, but I was not sure what exactly to call it. I will come up with a better name and I think that will solve most of the problems in this PR.

Opeyem1a (approved):
lgtm

PR #134, opened 2023-10-11
revamp priority algorithm and runner

Opeyem1a:
adding priority to the newer structure, early prep for what the algorithm runner will look like

### DIFF
--- a/api/ai/new/priority_algorithm/priority_algorithm.py
+++ b/api/ai/new/priority_algorithm/priority_algorithm.py
@@ -1,0 +1,175 @@
+import time
+from typing import cast, Dict, List
+
+from api.ai.new.interfaces.algorithm import Algorithm
+from api.ai.new.interfaces.algorithm_config import PriorityAlgorithmConfig
+from api.ai.new.interfaces.algorithm_options import (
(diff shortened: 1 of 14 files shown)


api/ai/new/priority_algorithm/priority_algorithm.py
```diff
@@ -0,0 +1,174 @@
...
+            iteration += 1
+
+        # the first team set is the "best" one
+        return self._unpack_priority_team_set(team_sets[0])
+
+    def mutate(self, team_set: PriorityTeamSet) -> List[PriorityTeamSet]:
+        """
+        Mutate a single teamset into child teamsets
+        """
+        algorithm = 1
```
justino599, line 96: I think this should be a part of the `PriorityAlgorithmConfig`

justino599 (changes requested):

api/ai/new/priority_algorithm/priority_algorithm.py
Opeyem1a: Oh yeah I gotta do the full refactor for it though

justino599 (approved):

PR #143, opened 2023-10-12
Add Multiple Round Robin With Adjusted Winner

### LINKED ISSUE #142: Multiple Round Robin with Adjusted Winners (MRR)

ketphan02:
### Sample run
```text
Class size: 20
-----
Team 2: 6 2 14 9 
-----
Team 4: 4 3 0 17 
-----
Team 1: 8 16 11 
-----
Team 6: 19 5 
-----
Team 5: 15 10 13 
-----
Team 3: 18 7 1 12 
```

Closes #142


api/ai/new/interfaces/algorithm_options.py
```diff
@@ -72,3 +72,9 @@ def validate(self):
 class PathGaspAlgorithmOptions(AlgorithmOptions):
     def validate(self):
         super().validate()
+
+
+@dataclass
+class MultipleRoundRobinAlgorithmOptions(AlgorithmOptions):
```
Opeyem1a, line 78: new algorithms will need to be added AlgorithmType enum

Opeyem1a (changes requested):

api/ai/new/multiple_round_robin_with_adjusted_winner/utils.py
```diff
@@ -0,0 +1,29 @@
...
+def calculate_value(student: Student, requirements: List[ProjectRequirement]) -> int:
+    return sum(
+        [
+            1 if student.meets_requirement(requirement) else -1
+            for requirement in requirements
+        ]
+    )
+
+
+def is_oef1(team_i: "TeamWithValues", team_j: "TeamWithValues") -> bool:
```
Opeyem1a, line 16: nit:
this function name is gross lmao

api/ai/new/multiple_round_robin_with_adjusted_winner/multiple_round_robin.py
```diff
@@ -0,0 +1,204 @@
...
+        self.projects = projects
+
+    @staticmethod
+    def _get_values_heap(
+        students: List[Student], teams: List[TeamWithValues]
+    ) -> Dict[int, List[StudentProjectValue]]:
+        """
+        Calculate each student values to each project and push them into a heap
+
+        This run in O(N * H)
```
Opeyem1a, line 66: A comment from the other PR is why do we need to annotate the big O everywhere? Like what does it add to the codebase to have this as a comment in the code?

api/ai/new/multiple_round_robin_with_adjusted_winner/multiple_round_robin.py
```diff
@@ -0,0 +1,204 @@
...
+        }
+        for team in teams:
+            project = team.project
+            for student in students:
+                new_value = StudentProjectValue(student, project)
+                heappush(values_heap[project.id], new_value)
+
+        return values_heap
+
+    @staticmethod
```
Opeyem1a, line 80: why make these methods static if you're just calling them with `self.` anyway?

If they are truly non-dependant on the `self.` then why aren't they utils?

(note: not every warning from pycharm means you need to fix it )

api/ai/new/multiple_round_robin_with_adjusted_winner/multiple_round_robin.py
```diff
@@ -0,0 +1,204 @@
...
+from api.models.student import Student
+from api.models.team_set import TeamSet
+
+
+class MultipleRoundRobinWithAdjustedWinnerAlgorithm(Algorithm):
+    def __init__(
+        self,
+        algorithm_options: MultipleRoundRobinAlgorithmOptions,
+        team_generation_options: TeamGenerationOptions,
+        projects: List[Project],
```
Opeyem1a, line 54: don't do this, if it's an argument of the algorithm then it should be in the `MultipleRoundRobinAlgorithmOptions`

Use the same casting technique from the other algorithms to cast it into the type you want

Opeyem1a (changes requested):

api/ai/new/multiple_round_robin_with_adjusted_winner/multiple_round_robin.py
ketphan02: It does not do much for the codebase, however, this will help a lot during the algorithm analysis process when formal documenting. If this is not preferred, I will add a TODO to remove it after formal documentation is established.

Opeyem1a: Make an issue to remove them later, but I'm cool with this line of logic for now.
tldr; it provides you value personally for now, so it can stay, but long term (after it is formally documented) I think it's pretty low value forother devs working on this project, so can be removed

api/ai/new/interfaces/algorithm_options.py
```diff
@@ -72,3 +73,11 @@ def validate(self):
...
+    projects: List[Project]
+
+    def validate(self):
+        super().validate()
```
Opeyem1a, line 83: you can do this as another ticket, but are there no constraints on the list of projects? like if it's empty does the algorithm run fine? if not you need to add those validations here

Opeyem1a (approved):
you made it soldier

PR #150, opened 2023-10-16
create singular simulation class

Opeyem1a:
- creates a class to run a single algorithm at a time (will soon include the support to pass a custom algorithm config)
- testing for this
- created SimulationSettings that this new Simulation class uses (will later switch BasicSimulationSet to use this)
- moved shared data in test directory to a `_data.py` file

PR #149, opened 2023-10-16
basic simulation set rename

Opeyem1a:
Makes more sense for this to be called BasicSimulationSet since it runs multiple algorithms at the same time
See the branch stacked on top of this for why this happened

PR #151, opened 2023-10-16
support custom configs on Simulation

### LINKED ISSUE #108: Develop a way to configure PriorityAlgorithm with different configs

Opeyem1a:
closes #108

### DIFF
--- a/tests/test_benchmarking/test_simulation/test_simulation.py
+++ b/tests/test_benchmarking/test_simulation/test_simulation.py
@@ -1,5 +1,6 @@
 import unittest
 
+from api.ai.new.interfaces.algorithm_config import PriorityAlgorithmConfig
 from api.models.enums import AlgorithmType
 from benchmarking.data.simulated_data.mock_student_provider import (
     MockStudentProvider,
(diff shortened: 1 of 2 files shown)


Opeyem1a:
* main
  * **PR #149** <a href="https://app.graphite.dev/github/pr/Teamable-Analytics/algorithms/149" target="_blank">Graphite</a> 
    * **PR #150** <a href="https://app.graphite.dev/github/pr/Teamable-Analytics/algorithms/150" target="_blank">Graphite</a> 
      * **PR #151** <a href="https://app.graphite.dev/github/pr/Teamable-Analytics/algorithms/151" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/Teamable-Analytics/algorithms/151?utm_source=stack-comment).

tests/test_benchmarking/test_simulation/test_simulation.py
```diff
@@ -58,3 +59,24 @@ def test_run__works_with_each_algorithm_type(self):
                 algorithm_type=algorithm_type,
                 settings=self.settings,
             ).run(num_runs=5)
+
+    def test_run__works_with_configs(self):
+        # fixme: probably some cool way to do this with python mocks
+        tracker = {"is_called": False}
+
+        def return_val(val):
```
Opeyem1a, line 67: wait this is dumb, I think it will always have called it just to resolve this value :'(

Hedgemon4 (approved):

tests/test_benchmarking/test_simulation/test_simulation.py
justino599: Yeah, not sure what you're trying to do here. From the title, it would seem that you just want to test that the constructor works given you've passed a config? But then why do you need this extra function?

Opeyem1a: I was trying to test that the config actually used is the correct config, essentially

So testing the actual simulation didn't use the default config

PR #158, opened 2023-10-17
Update boilerplate to make API runs

ketphan02:
### TLDR
Algorithm API is working now!
#### Major changes
- Updated `generate_teams_data_loader` to load students, algorithm options, algorithm type, team generation options
- Updated `generate_teams` to load use `GenerateTeamsInputData` to run the algorithm
- Add validations for requested json
- Added a `parse_json` method in the abstract class `AlgorithmOptions` to specify how to parse the algorithm options correctly. This resolve parsing and typing issues.
#### Minor changes
- Put the `manage.py` in the root directory, so it won't have issues with imports. Changed the config files accordingly.
- `Team` object needs a way to parse the data into dict, so added `todict` inside of `Team` (needs to remove later with the serializer)

### Sample request
```
{
    "students": [
        {
            "id": "1",
            "attributes": {
                "0": [1],
                "1": [3],
                "2": [0],
                "3": [1],
                "4": [0],
                "5": [3]
            },
            "relationships": {
                "2": "friend",
                "3": "enemy"
            },
            "project_preferences": [3, 4, 1]
        },
        {
            "id": "2",
            "attributes": {
                "0": [0],
                "1": [2],
                "2": [1],
                "3": [1],
                "4": [1],
                "5": [2]
            },
            "relationships": {
                "1": "friend"
            },
            "project_preferences": [1, 5, 2]
        },
        {
            "id": "3",
            "attributes": {
                "0": [2],
                "1": [1],
                "2": [2],
                "3": [1],
                "4": [0],
                "5": [3]
            },
            "relationships": {
                "1": "enemy",
                "2": "friend"
            },
            "project_preferences": [1, 2]
        }
    ],
    "algorithm_options": {
        "algorithm_type": "social",
        "max_project_preferences": 3
    },
    "team_generation_options": {
        "max_team_size": 3,
        "min_team_size": 2,
        "total_teams": 2,
        "initial_teams": []
    }
}
```
Output:
```
{
    "teams": {
        "num_teams": 2,
        "teams": [
            {
                "id": -1,
                "name": "Team 1",
                "student_ids": [
                    "1",
                    "2"
                ]
            },
            {
                "id": -1,
                "name": "Team 2",
                "student_ids": [
                    "3"
                ]
            }
        ]
    },
    "metadata": {
        "version": "v0.1.0",
        "timestamp": 1697528545.0011132
    },
    "error": null
}
```

### Further issue
- Teams does not have id (seeing team id in output is -1). In `algorithms/api/ai/new/interfaces/algorithm.py`, we have `# todo: unique id must be ensured now`
- Currently only support `Random`, `Social`, and `Weight` since `Priority` has not adapted to the new api structure


api/ai/new/social_algorithm/social_graph.py
```diff
@@ -56,5 +57,7 @@ def create_social_graph(students: List[Student]):
             social_graph[(student.id, other.id)] = Relationship.DEFAULT.value
             if other.id in student.relationships:
                 # update to real relationship value if one exists
-                social_graph[(student.id, other.id)] = student.relationships[other.id]
+                social_graph[(student.id, other.id)] = get_relationship_value(
```
Opeyem1a, line 60: I don't understand the need for introducing `get_relationship_value` in the actual algorithm code? 

I realize the error I made here (I think), but wouldn't it always be `social_graph[(student.id, other.id)] = student.relationships[other.id].value`
Can we not rely on Student.relationships to be Dict[str, Relationship]?

api/api/serializers/team_set.py
```diff
@@ -9,8 +9,11 @@ class TeamSetSerializer:
...
     def __init__(self, team_set: TeamSet):
-        pass
+        self.team_set = team_set
 
     @property
     def data(self) -> Dict[str, Any]:
-        pass
+        return {
+            "num_teams": self.team_set.num_teams,
+            "teams": [team.todict() for team in self.team_set.teams],
```
Opeyem1a, line 18: instead of team.todict(), there would just be a TeamSerializer, this is what Seth and Justin are working on

manage.py
Opeyem1a, line 1: why did we move this? I'm not sure I know what you mean by import issues

api/api/utils/generate_teams_data_loader.py
```diff
@@ -1,40 +1,76 @@
...
+        algorithm_option_class = get_algorithm_option_class(algorithm_type)
+        return algorithm_option_class.parse_json(algorithm_options)
+
+    def _validate_team_generation_options(
+        self, team_generation_options: Optional[Dict[str, Any]]
+    ):
+        if team_generation_options is None:
+            raise ValueError("team_generation_options is required")
+        Schema(int).validate(team_generation_options.get("total_teams"))
+        Schema(list).validate(team_generation_options.get("initial_teams"))
```
Opeyem1a, line 63: leave a todo or ticket because this validation is not as simple as just being a list

api/api/utils/generate_teams_data_loader.py
```diff
@@ -1,40 +1,76 @@
...
+        ]
 
     def _get_algorithm_options(self) -> AlgorithmOptions:
-        pass
+        algorithm_options = self.data.get("algorithm_options")
+        algorithm_type = AlgorithmType(algorithm_options.pop("algorithm_type"))
+        algorithm_option_class = get_algorithm_option_class(algorithm_type)
+        return algorithm_option_class.parse_json(algorithm_options)
+
+    def _validate_team_generation_options(
```
Opeyem1a, line 57: Is this function supposed to be here? I thought the validations were going to be in `validators/generate_teams_validator.py`?

Opeyem1a: The loader should have received validated data

api/api/utils/algorithm_option.py
```diff
@@ -0,0 +1,27 @@
...
+    RandomAlgorithmOptions,
+    WeightAlgorithmOptions,
+    SocialAlgorithmOptions,
+    PriorityAlgorithmOptions,
+    MultipleRoundRobinAlgorithmOptions,
+)
+from api.models.enums import AlgorithmType
+
+
+def get_algorithm_option_class(algorithm_type: AlgorithmType):
```
Opeyem1a, line 11: I feel like this method should be static on the AlgorithmRunner

api/api/utils/generate_teams_data_loader.py
```diff
@@ -1,40 +1,76 @@
...
+        Schema(int).validate(team_generation_options.get("min_team_size"))
 
     def _get_team_generation_options(self) -> TeamGenerationOptions:
-        pass
+        team_generation_options = self.data.get("team_generation_options")
+        self._validate_team_generation_options(team_generation_options)
+
+        return TeamGenerationOptions(
+            total_teams=team_generation_options.get("total_teams"),
+            initial_teams=team_generation_options.get("initial_teams"),
```
Opeyem1a, line 73: we'll also need to turn each thing in the initial_teams into objects here

api/api/utils/generate_teams_data_loader.py
```diff
@@ -1,40 +1,76 @@
...
-        pass
-
-    def _get_projects(self) -> List[Project]:
-        pass
+        students = self.data.get("students")
+        return [
+            Student(
+                _id=student.get("id"),
+                name=student.get("name"),
+                relationships=student.get("relationships"),
```
Opeyem1a, line 44: like why not just convert this dict to be `Dict[str, Relationship]` now and then we know the relationship values are always of the enum type going forward?

Opeyem1a: The loader's job is to make sure the data is consistent with the rest of the module

api/api/views/generate_teams.py
```diff
@@ -37,9 +36,8 @@ def generate_teams(self, request):
         input_data = GenerateTeamsDataLoader(request_data).load()
```
Opeyem1a, line 36: The line above (`GenerateTeamsValidator(request_data).validate()`) has gotta get done - can be in a separate PR though, let's take things chunk by chunk

api/api/utils/generate_teams_data_loader.py
```diff
@@ -1,40 +1,76 @@
...
+    algorithm_options: AlgorithmOptions
     team_generation_options: TeamGenerationOptions
-    projects: List[Project] = field(default_factory=list)
 
 
 class GenerateTeamsDataLoader:
     def __init__(self, validated_data: Dict[str, Any]):
         self.data = validated_data
 
     def load(self) -> GenerateTeamsInputData:
```
Opeyem1a, line 25: In general, if the loader is loading a non-primitive, it needs to convert that nested dict, list, or value into the correct enum or object type

Opeyem1a (changes requested):

api/api/serializers/team_set.py
ketphan02: I'll put down a TODO to update later

api/api/views/generate_teams.py
ketphan02: Oops

api/ai/algorithm_runner.py
```diff
@@ -54,3 +67,41 @@ def get_algorithm_from_type(algorithm_type: AlgorithmType):
...
+            return SocialAlgorithmOptions
+        if algorithm_type == AlgorithmType.PRIORITY:
+            return PriorityAlgorithmOptions
+        if algorithm_type == AlgorithmType.PRIORITY_NEW:
+            return PriorityAlgorithmOptions
+        if algorithm_type == AlgorithmType.MRR:
+            return MultipleRoundRobinAlgorithmOptions
+
+        raise NotImplementedError(
+            f"Algorithm type {algorithm_type} is not associated with an algorithm class!"
```
Opeyem1a, line 87: these error messages could be specific to the function

api/api/utils/generate_teams_data_loader.py
```diff
@@ -1,40 +1,84 @@
...
+        return algorithm_option_class.parse_json(algorithm_options)
 
     def _get_team_generation_options(self) -> TeamGenerationOptions:
-        pass
+        team_generation_options = self.data.get("team_generation_options")
+
+        return TeamGenerationOptions(
+            total_teams=team_generation_options.get("total_teams"),
+            initial_teams=[
+                TeamShell(team) for team in team_generation_options.get("initial_teams")
```
Opeyem1a, line 72: I'm not sure if `TeamShell(x)` does what you think? You might need `TeamShell(**team)` or something?

api/api/validators/generate_teams_validator.py
```diff
@@ -7,23 +14,59 @@ def validate(self):
...
         self.validate_algorithm_options()
-        self.validate_projects()
+        self.validate_algorithm_configs()
         self.validate_students()
 
     def _validate_schema(self):
-        pass
+        Schema(dict).validate(self.data)
+        Schema(list).validate(self.data.get("students"))
+        Schema(dict).validate(self.data.get("algorithm_options"))
```
Opeyem1a, line 23: this PR is doing a lot of things and I feel like some are falling through the cracks -> this structure is more complex than just being a dictionary

And others too -> students is not just a list of anything and so on and so on, if you're choosing to do all the vlaidation in this PR, it needs to be done properly

api/api/validators/generate_teams_validator.py
```diff
@@ -7,23 +14,59 @@ def validate(self):
...
+        algorithm_config = self.data.get("algorithm_config")
+        algorithm_config_cls = AlgorithmRunner.get_algorithm_config_class(
+            algorithm_type
+        )
 
-    def validate_projects(self):
-        pass
+        algorithm_config_cls(**algorithm_config).validate()
 
     def validate_students(self):
```
Opeyem1a, line 56: Not really the kind of validation we need in this method.

Creating the `Student` objects is the job of the loader, the validator just needs to make sure the data is all good to go.

For example, something important to validate is that every student dict has a unique id, or that the ids in their `project_preference` list actually correlate to real projects that were passed in

api/api/validators/generate_teams_validator.py
```diff
@@ -7,23 +14,59 @@ def validate(self):
...
 
     def validate_algorithm_options(self):
-        pass
+        algorithm_option = dict.copy(self.data.get("algorithm_options"))
+        algorithm_type = AlgorithmType(algorithm_option.pop("algorithm_type"))
+        algorithm_option_cls = AlgorithmRunner.get_algorithm_option_class(
+            algorithm_type
+        )
+
+        algorithm_option_cls(**algorithm_option).validate()
```
Opeyem1a, line 43: we shouldn't really be introducing data class inits at this level imo, this class is just a screen of the raw data to make sure it is able to be converted into the dataclasses properly by the loader. 

When the loader creates these objects it will run the validate() on the dataclass automatically so replicating that logic here doesn't grant us anything

Opeyem1a: this applies to the other methods you have here

api/api/validators/generate_teams_validator.py
Opeyem1a: We can chat about this if this feels confusing, I don't think I properly communicated the semantic difference between classes accurately

api/models/student.py
```diff
@@ -36,3 +38,14 @@ def meets_requirement(self, requirement: ProjectRequirement) -> bool:
             else:  # default case is RequirementOperator.EXACTLY
                 is_met |= value == requirement.value
         return is_met
+
+    def validate(self):
```
Opeyem1a, line 42: Like this is something that the validator class would do, I wouldn't put it here imo

Opeyem1a (changes requested):
I feel like we probably should still chat to really sort out the layers of validation and loading

Opeyem1a (approved):

PR #159, opened 2023-10-20
Implemented JSON Serializer for Students

Hedgemon4:
### Description
- implemented a serializer to encode and decode students into JSON
- moved the student model into a new folder containing the model and the serializer


tests/test_api/test_models/test_student/test_student_serializer.py
```diff
@@ -0,0 +1,52 @@
...
+                relationships={1: Relationship.FRIEND},
+                project_preferences=[3, 2],
+            ),
+        ]
+        cls.json_students = [
+            '{"_id": 1, "name": "Harry", "attributes": {"4": [2, 56, 1], "11": [2], "234": [-1, 0]}, "relationships": {"2": -1}, "project_preferences": [6, 1]}',
+            '{"_id": 2, "name": "Jane", "attributes": {"4": [31, 2], "11": [], "234": [1, -1]}, "relationships": {"1": -1}, "project_preferences": [3, 2]}',
+        ]
+
+    def test_student_serializer__encode_returns_string(self):
```
Hedgemon4, line 34: I don't know if we need this method. It does not actually test what encode returns because encode returns a dict. It checks that the returned dict gets turned into a string. Let me know what you think.

Opeyem1a: Yeah I'd say not super needed here tbh

Opeyem1a (approved):
Seems gucci, we might want to start making modules so the imports don't just infinitely become more gross though - separate PR type thing we can talk about a different day

Opeyem1a (approved):
lgtm

PR #161, opened 2023-10-22
Implement Team Serializer

Hedgemon4:
Description

- refactored project requirement and student serializer to allow for chain calls to decode 
- Added serializer for teams

PR #164, opened 2023-10-24
Update Simulation to check for and write to cache

### LINKED ISSUE #163: Update Simulation to check for and write to cache
### Goals
- Add a "cache key" as a parameter to `Simulation`. Is more or less a filename that data should be saved to.
- Data should be saved as JSON with 2 main keys: `metadata`, and `team_sets`
    - `metadata` includes timestamp, the hash of the git commit on which the cache file was generated, and other stuff that is not the teamsets (TBD)
    - `team_sets` is just the JSON serialized version of the simulation output teams (Being worked on in #161)
- Caches are saved to root level `cache` directory

justino599:
### Changes
- [x] Add a "cache key" as a parameter to `Simulation`. Is more or less a filename that data should be saved to.
- [x] Data should be saved as JSON with 3 main keys: `metadata`, `team_sets`, and `runtimes`
    - [x] `metadata` includes timestamp, the hash of the git commit on which the cache file was generated, and other stuff that is not the teamsets (TBD)
    - [x] `team_sets` is just the JSON serialized version of the simulation output teams (Being worked on in #161)
    - [x] `runtimes` is a list of the amount of time it took to generate the `TeamSet` at the same index in `team_sets`
- [x] Caches are saved to root level `simulation_cache` directory

### Example
The code
```python
mock_simluation_result: List[TeamSet] = {...}
mock_runtimes: List[float] = [1.8, 2.0]
cache = SimulationCache("test_cache")
cache.save(mock_simulation_result, mock_runtimes)
```

produces the cache

```json
{"metadata": {"timestamp": 1698387131.742896, "commit_hash": "ea4b483735ef68440c37b435a3f8321d5582a436"}, "team_sets": [{"_id": 0, "name": "TeamSet0", "teams": [{"_id": 0, "name": "Team0", "project_id": null, "requirements": [], "students": [{"_id": 0, "name": null, "attributes": {}, "relationships": {}, "project_preferences": []}, {"_id": 1, "name": null, "attributes": {}, "relationships": {}, "project_preferences": []}]}, {"_id": 1, "name": "Team1", "project_id": null, "requirements": [], "students": [{"_id": 2, "name": null, "attributes": {}, "relationships": {}, "project_preferences": []}, {"_id": 3, "name": null, "attributes": {}, "relationships": {}, "project_preferences": []}]}]}, {"_id": 1, "name": "TeamSet1", "teams": [{"_id": 0, "name": "Team0", "project_id": null, "requirements": [], "students": [{"_id": 0, "name": null, "attributes": {}, "relationships": {}, "project_preferences": []}, {"_id": 2, "name": null, "attributes": {}, "relationships": {}, "project_preferences": []}]}, {"_id": 1, "name": "Team1", "project_id": null, "requirements": [], "students": [{"_id": 1, "name": null, "attributes": {}, "relationships": {}, "project_preferences": []}, {"_id": 3, "name": null, "attributes": {}, "relationships": {}, "project_preferences": []}]}]}], "runtimes": [1.8, 2.0]}
```

Closes #163


benchmarking/caching/simulation_cache.py
```diff
@@ -0,0 +1,115 @@
...
+class SimulationCache:
+    """
+    This class is used to cache the results of a simulation, which is a List[TeamSet], where each element of the list is a different run.
+    All TeamSets should have been generated by the same algorithm (with the same settings).
+    It uses the cache_key to generate a file with the results of the simulation.
+    """
+
+    def __init__(self, cache_key: str) -> None:
+        self.cache_key = cache_key
+        self._data = None
```
justino599, line 19: I'm not a fan not being able to type this, but if I do, it generates a warning about assigning None to type Dict[str, Any]

benchmarking/caching/simulation_cache.py
```diff
@@ -0,0 +1,115 @@
...
+        Overwrites any existing cache with the same cache_key.
+        """
+
+        # Get metadata
+        metadata = metadata or {}
+        metadata["timestamp"] = datetime.now().isoformat()
+        # Get latest commit hash. This is so that we can track which commit generated the cache and go back to the code that generated it if needed. Would then be accessible at https://github.com/Teamable-Analytics/algorithms/commit/<commit_hash>
+        metadata["commit_hash"] = git.Repo(
+            search_parent_directories=True
+        ).head.object.hexsha
```
justino599, line 64: These hashes are actually independent of the branch they're on, so this works on both main and any other branch

benchmarking/simulation/simulation.py
```diff
@@ -33,6 +34,14 @@ def __init__(
         self.team_sets = []
 
     def run(self, num_runs: int) -> SimulationArtifact:
+        if self.settings.cache_key:
+            cache = SimulationCache(self.settings.cache_key)
+            if cache.exists():
+                self.team_sets = cache.get_teams()
+                self.run_times = cache.get_runtimes()
+                # TODO: If len(self.team_sets) < num_runs, run the remaining runs
```
Opeyem1a, line 42: Make a separate ticket for this, I think we include this in the start/stop functionality

benchmarking/simulation/simulation.py
```diff
@@ -67,4 +76,9 @@ def run(self, num_runs: int) -> SimulationArtifact:
             self.run_times.append(end_time - start_time)
             self.team_sets.append(team_set)
 
+            # Save result to cache. Do this inside the loop so that if the program crashes, we still have the results from the previous runs.
+            if self.settings.cache_key:
```
Opeyem1a, line 80: probably include a stagger to this -> grab the counter for the loop (I guess give it a name now) and do a `if i % SimulationCache.CACHE_EVERY_N_TRIALS` and add this constant to SimulationCache 

Kinda just spitballing though, put the constant wherever makes the most sense and add the logic where makes the most sense too.

More complex idea is to have a method in SimulationCache that's like `add_record(self, team_set: TeamSet, run_time: float)` and then inside that method, it saves to a dict it's keeping track of internally and chooses to do an append to the file when it decides it's worth doing, but this is kinda more thoughts and something to consider after perhaps

benchmarking/simulation/simulation.py
```diff
@@ -33,6 +34,14 @@ def __init__(
         self.team_sets = []
 
     def run(self, num_runs: int) -> SimulationArtifact:
+        if self.settings.cache_key:
```
Opeyem1a, line 37: maybe make a ticket for the whole `cache_key="folder/folder/name` where it would make those nested folders automatically

tests/test_benchmarking/test_caching/test_simulation_cache.py
```diff
@@ -0,0 +1,256 @@
...
+mock_runtimes: List[float] = [
+    1.8,
+    2.0,
+]
+
+
+class TestSimulationCache(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        # I'm paranoid about these tests modifying the cache, so move the cache to a different location as a precaution
```
Opeyem1a, line 70: something you also might be able to do is just mock a bunch of the `os` methods and have it "make files" and "open files" from somewhere else instead. 

it's a bit sus to do a full file copy tbh 😭 so I think we should explore an approach that uses mocks to still test core functionality 

i.e. if you mock the response of `path.dirname(__file__)`, you can make it think you are anywhere you want in the file structure

In other cases, you can just mock a call to the os function that deletes files and just make sure it was called, instead of checking that an actual file was actually deleted.

You could also create a `prepare_cache_data` method in SimulationCache that returns the exact json that will be saved, so you can unit test that fields are in that dict and have the only thing that actually needs to check if it created a file be the single "save_data" method or whatever

I'm trying to drop ideas but feel free to call me to chat about this - I think doing shenanigan directory moving is a scary thing

Opeyem1a (changes requested):

benchmarking/simulation/simulation.py
justino599: Added the add_run() method, but not adding the `CACHE_EVERY_N_TRIALS` as performance is not a worry here

tests/test_benchmarking/test_caching/test_simulation_cache.py
justino599: Took me a while to figure out how to mock stuff, but it works well

Opeyem1a (approved):
solid

PR #171, opened 2023-10-28
Nested cache directories

### LINKED ISSUE #170: Cache key can contain "/" which will create store in a seperate directory
### Issue
Cache keys should be able to contain "/" such as `priority/local_max/30_iterations` which would create a cache at `/run_cache/priority/local_max/30_iterations.json`, for example

justino599:
### Changes
- Actually not much, just make sure the directory is created and normalize path

Closes #170


Opeyem1a (approved):
lgtm

PR #201, opened 2023-11-07
Add API validation

ketphan02:
- API validation for:
  - `students`
  - `team_generation_options`
  - `algorithm_type`

Since `team_generation_options` is too big, we will have a separate PR for it

### DIFF
--- a/api/api/utils/generate_teams_data_loader.py
+++ b/api/api/utils/generate_teams_data_loader.py
@@ -40,10 +40,10 @@ class GenerateTeamsDataLoader:
         students = self.data.get("students")
         return [
             Student(
-                _id=student.get("id"),
+                _id=int(student.get("id")),
                 name=student.get("name"),
                 relationships={
-                    student_id: get_relationship(relationship)
+                    int(student_id): get_relationship(relationship)
                     for student_id, relationship in student.get("relationships").items()
                 },
                 attributes=student.get("attributes"),
@@ -67,7 +67,12 @@ class GenerateTeamsDataLoader:
         return TeamGenerationOptions(
             total_teams=team_generation_options.get("total_teams"),
             initial_teams=[
-                TeamShell(**team)
+                TeamShell(
+                    _id=int(team.get("id")),
+                    name=team.get("name"),
+                    project_id=int(team.get("project_id")),
+                    requirements=team.get("requirements", []),
+                )
                 for team in team_generation_options.get("initial_teams")
             ],
             max_team_size=team_generation_options.get("max_team_size"),
--- a/api/api/utils/relationship.py
+++ b/api/api/utils/relationship.py
@@ -9,3 +9,13 @@ def get_relationship(relationship: str) -> Relationship:
     if relationship == "default":
         return Relationship.DEFAULT
     raise ValueError(f"Relationship {relationship} is not a valid relationship")
+
+
+def get_relationship_str(relationship: Relationship) -> str:
+    if relationship == Relationship.FRIEND:
+        return "friend"
+    if relationship == Relationship.ENEMY:
+        return "enemy"
+    if relationship == Relationship.DEFAULT:
+        return "default"
+    raise ValueError(f"Relationship {relationship} is not a valid relationship")
--- a/api/api/validators/generate_teams_validator.py
+++ b/api/api/validators/generate_teams_validator.py
@@ -116,7 +116,7 @@ class GenerateTeamsValidator(Validator):
         all_attributes = set()
         for team in teams:
             all_attributes.update(
-                [requirement.attribute for requirement in team.get("requirements")] if team.get("requirements") else []
+                [requirement.get("attribute") for requirement in team.get("requirements")] if team.get("requirements") else []
             )
         for student in students:
             attributes = student.get("attributes")
@@ -134,7 +134,7 @@ class GenerateTeamsValidator(Validator):
             # Validate if attribute keys exist
             if not all_attributes.issuperset(attribute_keys):
                 raise SchemaError(
-                    f"Student {student.get('id')} has attributes that do not exist"
+                    f"Student {student.get('id')} has attributes that do not exist."
                 )
 
         for student in students:
--- a/api/api/validators/generate_teams_validator.py
+++ b/api/api/validators/generate_teams_validator.py
@@ -25,11 +25,13 @@ class GenerateTeamsValidator(Validator):
         self.validate_students()
 
     def _validate_schema(self):
-        Schema({
-            "algorithm_options": Dict,
-            "students": List[Dict],
-            "team_generation_options": Dict,
-        }).validate(self.data)
+        Schema(
+            {
+                "algorithm_options": Dict,
+                "students": List[Dict],
+                "team_generation_options": Dict,
+            }
+        ).validate(self.data)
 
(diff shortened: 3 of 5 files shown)


api/api/utils/relationship.py
```diff
@@ -9,3 +9,13 @@ def get_relationship(relationship: str) -> Relationship:
     if relationship == "default":
         return Relationship.DEFAULT
     raise ValueError(f"Relationship {relationship} is not a valid relationship")
+
+
+def get_relationship_str(relationship: Relationship) -> str:
```
Opeyem1a, line 14: idk if we need this? Why not just `str(relationship)` instead of a util function for it?

api/api/utils/generate_teams_data_loader.py
```diff
@@ -40,10 +40,10 @@ def _get_students(self) -> List[Student]:
         students = self.data.get("students")
         return [
             Student(
```
Opeyem1a, line 42: now that I think about it, couldn't we use the serializer here?

api/api/validators/generate_teams_validator.py
```diff
@@ -49,31 +52,148 @@ def validate_algorithm_options(self):
...
-        #
-        #     }
-        # ]).validate(self.data)
+        students: List = self.data.get("students")
+        Schema(
+            [
+                {
+                    "id": int,
+                    Optional("name"): str,
+                    Optional("attributes"): {str: [int]},
```
Opeyem1a, line 61: each of the attribute keys technically need to be able to be parsed into ints, not sure if we can do this with `Schema` but would be cool

api/api/validators/generate_teams_validator.py
```diff
@@ -49,31 +52,148 @@ def validate_algorithm_options(self):
...
+                    Optional("relationships"): {
+                        str: Or(
```
Opeyem1a, line 63: same here with the string technically being an int

api/api/utils/relationship.py
Opeyem1a: oh because its a float, hmm

api/api/validators/generate_teams_validator.py
```diff
@@ -49,31 +52,148 @@ def validate_algorithm_options(self):
...
+                            ]
+                        )
+                    },
+                    Optional("project_preferences"): [int],
+                }
+            ]
+        ).validate(students)
 
-        pass
+        student_ids = set()
```
Opeyem1a, line 75: for testing purposes as well as readability, I think it might be worth putting each "validation" in a separate function of this class, and just calling them where needed.

i.e. `self.validate_students()` could call `self.validate_student_ids_unique()` and `self.validate_students_project_preferences()`, etc

api/api/validators/generate_teams_validator.py
```diff
@@ -49,31 +52,148 @@ def validate_algorithm_options(self):
...
-        pass
+        team_options = self.data.get("team_generation_options")
+
+        Schema(
+            {
+                "initial_teams": [
+                    {
+                        "id": int,
+                        Optional("name"): str,
+                        "project_id": int,
```
Opeyem1a, line 178: this is also optional

api/api/validators/generate_teams_validator.py
```diff
@@ -49,31 +52,148 @@ def validate_algorithm_options(self):
...
+                        Optional("requirements"): [
+                            {
+                                "attribute": int,
+                                "operator": str,
```
Opeyem1a, line 182: there is a discrete number of possible operators

tests/test_api/test_validators/test_generate_teams_validator/test_generate_teams_validator.py
```diff
@@ -0,0 +1,219 @@
...
+            validator = GenerateTeamsValidator(self.valid_data)
+            validator.validate()
+        except SchemaError as e:
+            self.fail("Valid data raised SchemaError unexpectedly")
+
+    def test_errors__top_levels(self):
+        with self.assertRaises(SchemaError) as e:
+            validator = GenerateTeamsValidator(self.missing_students_data)
+            validator.validate()
+        self.assertEqual(e.exception.code, "Missing key: 'students'")
```
Opeyem1a, line 172: don't remove now, but for the future idk if we should write tests that are this specific, but I know why you did it.

I feel like if you broke the validator class into smaller methods, the tests could also be smaller, and then you wouldnt need to test the exact match of the exception code, you could just test if individual validation methods raise errors

Opeyem1a (changes requested):

api/api/utils/generate_teams_data_loader.py
ketphan02: Another PR then

api/api/validators/generate_teams_validator.py
ketphan02: line 136 checked that

api/api/validators/generate_teams_validator.py
ketphan02: line 136 checked that

api/api/validators/generate_teams_validator.py
```diff
@@ -44,32 +50,167 @@ def validate_algorithm_options(self):
...
+                raise SchemaError(f"Student {student.get('id')} has no attributes.")
+
+            # Validate if attribute keys integer string
+            try:
+                attribute_keys = list(map(int, attributes.keys())) if attributes else []
+            except ValueError:
+                raise SchemaError("Attribute keys must be integers.")
+
+            # Validate if attribute keys are unique
+            if len(attribute_keys) != len(set(attribute_keys)):
```
Opeyem1a, line 115: it's impossible to have the same key twice in a dict (hashmap) - my bad if I told you to do this, because on second throught it doesn't need to be checked

api/api/validators/generate_teams_validator.py
```diff
@@ -44,32 +50,167 @@ def validate_algorithm_options(self):
...
+                raise SchemaError("Attribute keys must be unique.")
+
+            # Validate if attribute keys exist
+            if not all_attributes.issuperset(attribute_keys):
```
Opeyem1a, line 119: On second thought here too, it's cool if students have more attribute keys than the project requirements specify, no logical issue there. It's only minorly an issue if project ids have requirements for attributes that no student has, but I don't think we need to check that. It causes no problems for our system, just a wonky team generation for the user (i.e. user error)

tldr; remove

api/api/validators/generate_teams_validator.py
```diff
@@ -44,32 +50,167 @@ def validate_algorithm_options(self):
...
+            relationships = student.get("relationships")
+
+            # Validate if relationship keys integer string
+            try:
+                relationship_keys = list(map(int, relationships.keys()))
+            except ValueError:
+                raise SchemaError("Relationship keys must be integers.")
+
+            # Validate if relationship keys are unique
+            if len(relationship_keys) != len(set(relationship_keys)):
```
Opeyem1a, line 137: unneeded check, keys of dictionaries are unique by definition

Opeyem1a (changes requested):
almost there - just some quick things to remove

Opeyem1a (approved):
we have awaited this moment 🥺💪

PR #203, opened 2023-11-07
Add api validation: algorithm_generation_options

ketphan02:
Part 2 of #201


ketphan02:
Reminder: change base to `main` before merge

api/ai/new/interfaces/algorithm_options.py
```diff
@@ -42,6 +48,10 @@ def validate(self):
     def parse_json(_: Dict[str, Any]) -> "RandomAlgorithmOptions":
         return RandomAlgorithmOptions()
 
+    @staticmethod
+    def get_schema() -> Schema:
+        return Schema({"algorithm_type": Const(AlgorithmType.RANDOM.value)})
```
Opeyem1a, line 53: it's BRUTAL that we have to put the algorithm type in the schema even though it's not in the dataclass, if we can avoid this it would be sick

api/ai/new/interfaces/algorithm_options.py
```diff
@@ -141,12 +172,68 @@ def parse_json(json: Dict[str, Any]) -> "PriorityAlgorithmOptions":
...
+                SchemaOptional("social_weight"): int,
+                SchemaOptional("diversity_weight"): int,
+                SchemaOptional("preference_weight"): int,
+                SchemaOptional("friend_behaviour"): Or(
+                    *[behaviour.value for behaviour in RelationshipBehaviour]
+                ),
+                SchemaOptional("enemy_behaviour"): Or(
+                    *[behaviour.value for behaviour in RelationshipBehaviour]
+                ),
+                SchemaOptional("attributes_to_diversify"): Optional[List[int]],
```
Opeyem1a, line 205: why optional list int?

api/ai/new/interfaces/algorithm_options.py
```diff
@@ -141,12 +172,68 @@ def parse_json(json: Dict[str, Any]) -> "PriorityAlgorithmOptions":
             max_project_preferences=max_project_preferences,
         )
 
+    @staticmethod
+    def get_schema() -> Schema:
+        return Schema(
+            {
+                "algorithm_type": Const(AlgorithmType.PRIORITY_NEW.value),
```
Opeyem1a, line 179: use PRIORITY here, i'm gonna get rid of PRIORITY_NEW

api/ai/new/priority_algorithm/priority/priority.py
```diff
@@ -35,7 +35,8 @@ def validate(self):
     def satisfied_by(self, students: List[Student]) -> bool:
         count = 0
         for student in students:
-            count += self.value in student.attributes.get(self.attribute_id, [])
+            if student.attributes is not None:
```
Opeyem1a, line 38: student.attributes should never be None, if it is, something massive is wrong and we *want* it to throw an error

tests/test_api/test_validators/test_generate_teams_validator/test_generate_teams_validator.py
```diff
@@ -217,3 +267,45 @@ def test_errors__team_generation_options(self):
...
+        )
+        invalid_field_random_algorithm_data["algorithm_options"][
+            "max_project_preferences"
+        ] = 3
+        with self.assertRaises(SchemaError) as e:
+            validator = GenerateTeamsValidator(invalid_field_random_algorithm_data)
+            validator.validate()
+        self.assertIn("Wrong key 'max_project_preferences' in ", e.exception.code)
+
+    def test_errors__algorithm_options__weight_algorithm(self):
```
Opeyem1a, line 283: idk if I like this naming scheme, I think the "test_{name of function}__{explain test briefly}" is a better idea here still

Opeyem1a (changes requested):

tests/test_api/test_validators/test_generate_teams_validator/test_generate_teams_validator.py
Opeyem1a: kind of a nitpick but also I think we should have test names just give context on what the problem is without having to read the whole test

api/ai/priority_algorithm/priority/priority.py
```diff
@@ -37,6 +37,9 @@ def satisfaction(self, students: List[Student]) -> float:
     def satisfied_by(self, students: List[Student]) -> bool:
         count = 0
         for student in students:
+            if student.attributes is None:
```
Opeyem1a, line 40: it's never None, this should be impossible and if it happens, I want it to throw an error

api/api/validators/generate_teams_validator.py
```diff
@@ -44,11 +45,15 @@ def validate_algorithm_type(self):
             )
 
     def validate_algorithm_options(self):
-        # todo: validate that anything that is an id is real (?)
-        #   so attributes_to_concentrate is a list of attribute ids, ideally these exist on each student
+        # TODO: Do not commit this
```
Opeyem1a, line 48: prolly dont commit this

Opeyem1a (changes requested):
some more small things to remove

Opeyem1a (approved):
lgtm

PR #207, opened 2023-11-09
Implement Double Random Local Max

Hedgemon4:
### Description
- implemented a version of local max where both teams to optimize are chosen randomly


api/ai/new/priority_algorithm/mutations/utils.py
```diff
@@ -26,3 +26,13 @@ def score(
             )
         ]
     )
+
+
+def get_available_priority_teams(
```
Opeyem1a, line 31: wisdom

api/ai/new/priority_algorithm/mutations/local_max.py
```diff
@@ -75,6 +70,29 @@ def mutate_local_max_random(
...
+    combinations of students on those teams. It replaces one of the teams with the team with the highest score found,
+    and puts all other students on the other team. This implementation does not account for if the second team's score
+    increases or decreases.
+    """
+    available_priority_teams = get_available_priority_teams(priority_team_set)
+    try:
+        if len(available_priority_teams) < 2:
+            return priority_team_set
+        team_1, team_2 = random.sample(available_priority_teams, 2)
+        local_max(team_1, team_2, priorities, student_dict)
```
Opeyem1a, line 89: oh it performs the switches in-place?

Only concern is what if it like removes a student, somehow hits and error and then the catch statement just returns that malformed team set? Is this possible? Not a concern if not

Opeyem1a (approved):
just a question, if all good then good to merge

api/ai/new/priority_algorithm/mutations/local_max.py
Hedgemon4: It does perform the switch in place. I am not sure how this would happen, but I think it could be better to throw an error anyways. Not really sure it makes sense to catch this value error. If there is a problem, we probably want to know.

Opeyem1a: Yeah I would agree that we should just let the error be thrown

PR #208, opened 2023-11-11
implement diversity priority without tokenization [204]

### LINKED ISSUE #204: Add new type of priority - regular diversification
The priority algorithm currently only supports tokenization constraints (i.e. diversifying/concentrating with min of or max of X), but we should also support normal diversify and concentrate priorities.

Users should be able to say "diversify gender as a top priority" without needing to specify a tokenization constraint with that.

The resolution of this issue also implicitly includes the structure to support varied kinds of priorities in general (e.g. "make sure project requirements are met", "make sure friends are together")

Opeyem1a:
closes #204 

- Adds regular diversification/concentration priorities
    - implemented into the algorithm runner so it'll auto convert
- The thing to note here is that Priority subclasses will now have a `def satisfaction() -> float` instead of a `def satisfied_by() -> bool`. This is because the logic actually works essentially the same with minor modifications and then allows the question "does this group of students satisfy this priority" to be more like "how much does this team of students satisfy this priority"

### DIFF
--- a/api/ai/new/priority_algorithm/mutations/robinhood.py
+++ b/api/ai/new/priority_algorithm/mutations/robinhood.py
@@ -8,6 +8,8 @@ from api.ai.new.priority_algorithm.custom_models import PriorityTeamSet, Priorit
 from api.models.student import Student
 
 
+ROBINHOOD_SATISFACTION_THRESHOLD = 0.8
+
 def mutate_robinhood(
     priority_team_set: PriorityTeamSet,
     priorities: List[Priority],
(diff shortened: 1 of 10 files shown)


Opeyem1a:
* main
  * **PR #208** <a href="https://app.graphite.dev/github/pr/Teamable-Analytics/algorithms/208?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #209** <a href="https://app.graphite.dev/github/pr/Teamable-Analytics/algorithms/209?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/Teamable-Analytics/algorithms/208?utm_source=stack-comment).

PR #229, opened 2023-11-15
remove match_skills from requirement priority

Opeyem1a:
Unnecessary, if you don't want a requirement goal to match skills then you simply wouldn't include a requirement goal


Hedgemon4 (approved):
LGTM

PR #230, opened 2023-11-16
Envy-freeness up to 1 item (EF1) metric

ketphan02:


benchmarking/evaluations/metrics/envy_free_up_to_one_item.py
```diff
@@ -0,0 +1,66 @@
...
+class EnvyFreenessUpToOneItem(TeamSetMetric):
+    """
+    Calculate the envy-free up to one metric for the given attribute for each team and average the results
+
+    A allocation is envy-free up to one if no agent envies another agent by more than one unit of the given attribute
+    Meaning if A is envying B, then A will no longer envy B if we remove one unit of the given attribute from  B
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(name="EnvyFreenessUpToOneItem", *args, **kwargs)
```
Opeyem1a, line 17: if it has a theoretical range, include that here
Don't set the name here, it is intended that people can overwrite the name as needed

benchmarking/evaluations/metrics/envy_free_up_to_one_item.py
```diff
@@ -0,0 +1,66 @@
+import copy
+
+from api.models.team import Team
+from api.models.team_set import TeamSet
+from benchmarking.evaluations.interfaces import TeamSetMetric
+
+
+class EnvyFreenessUpToOneItem(TeamSetMetric):
```
Opeyem1a, line 8: I feel like even after reading this name and the comment, I still don't understand what this is calculating conceptually

I know this is something from a paper, but it's good to keep in mind that the code should not be dictated by the specific naming conventions and terminology of papers -- and if it actually is a non-generalized metric that only applies for the specific algorithms from other papers, this will devolve into the kind of function destined for "never use this again once this course ends"

Opeyem1a (changes requested):

benchmarking/evaluations/metrics/envy_free_up_to_one_item.py
```diff
@@ -0,0 +1,76 @@
...
+
+    Definitions:
+
+    - Envy: Team A envies Team B if Team A's utility is less than Team B's utility
+    - Envy-free up to one item (ef1): Team B ef1 Team A if Team B envies Team A and if Team A removes one student, Team B no longer envies Team A
+    """
+
+    def __init__(
+        self,
+        calculate_utilities: Callable[[TeamSet], Dict[int, Dict[int]]],
```
Opeyem1a, line 23: i don't understand this type?

You only call calculate utilities with team set, so what is this second parameter supposed to be?

Also, is it wise to calculate the utilities of all teams at once? I feel like this is more complicated than having the lambda just be a function that takes a student and a team and returns a number, does this work for your use cases?

Opeyem1a (changes requested):
few more questions

Opeyem1a (approved):
Seems fine

PR #231, opened 2023-11-16
Benchmark Priority Algorithm Parameters with Diversify Gender Min 2 Scenario

Hedgemon4:
### Description
- created benchmark file for each parameter of the priority algorithm for diversifying gender with min 2 2-scenario


benchmarking/runs/priority_algorithm/diversify_gender_min_2/class_size.py
```diff
@@ -0,0 +1,244 @@
...
+from benchmarking.runs.interfaces import Run
+from benchmarking.simulation.goal_to_priority import goals_to_priorities
+from benchmarking.simulation.insight import Insight
+from benchmarking.simulation.simulation_set import SimulationSet, SimulationSetArtifact
+from benchmarking.simulation.simulation_settings import SimulationSettings
+
+
+class DiversifyGenderMin2ClassSize(Run):
+    @staticmethod
+    def start(num_trials: int = 10, generate_graphs: bool = False):
```
Opeyem1a, line 46: I think generate_graphs should default to true

benchmarking/runs/priority_algorithm/diversify_gender_min_2/max_keep.py
```diff
@@ -0,0 +1,240 @@
...
+    @staticmethod
+    def start(num_trials: int = 10, generate_graphs: bool = False):
+        """
+        Goal: Run diversify gender scenario while varying the maximum keep argument for the priority algorithm.
+        """
+
+        # Defining our changing x-values (in the graph sense)
+        max_keep = list(range(1, 10))
+        ratio_of_female_students = 0.4
+        max_iterate = 300
```
Opeyem1a, line 53: these things being held constant are kinda super arbitrary, how did we arrive at them?

benchmarking/runs/priority_algorithm/diversify_gender_min_2/max_spread.py
```diff
@@ -0,0 +1,243 @@
...
+                            ],
+                            MAX_TIME=max_time,
+                            MAX_ITERATE=max_iterate,
+                            MAX_KEEP=max_keep,
+                            MAX_SPREAD=spread,
+                        ),
+                        PriorityAlgorithmConfig(
+                            name="local_max_double_random",
+                            MUTATIONS=[
+                                (mutate_local_max_double_random, 1),
```
Opeyem1a, line 138: things get weird here because if we have max spread of 10, for example, why only run the local max double random once? Why not 5 times?

benchmarking/runs/priority_algorithm/diversify_gender_min_2/max_time.py
```diff
@@ -0,0 +1,240 @@
...
+    @staticmethod
+    def start(num_trials: int = 10, generate_graphs: bool = False):
+        """
+        Goal: Run diversify gender scenario while varying the maximum time argument for the priority algorithm.
+        """
+
+        # Defining our changing x-values (in the graph sense)
+        times = list(range(2, 21, 2))
+        ratio_of_female_students = 0.4
+        max_iterate = 300
```
Opeyem1a, line 53: In general, if this is a group of runs, I think you should store these constants in one place in the directory - like all the "when we are not varying this thing, this is the value we use should be in once place and referenced, it feels spooky to have them just be defined everywhere like this

I think it also helps readability because down below in the configs it looks like everything is a variable, when in reality only 1 thing changes and the rest could be `CONSTANT_THIS_THING`

benchmarking/runs/priority_algorithm/diversify_gender_min_2/num_iterations.py
```diff
@@ -0,0 +1,242 @@
...
+        graph_runtime_dict = {}
+        graph_avg_gini_dict = {}
+        graph_priority_dict = {}
+        graph_dicts = [
+            graph_runtime_dict,
+            graph_avg_gini_dict,
+            graph_priority_dict,
+        ]
+
+        metrics: Dict[str, TeamSetMetric] = {
```
Opeyem1a, line 70: I also think if you have this group of runs, you can just put all this repeated logic in one place, like defining the metrics and running all the graphs, it's the same for each right?

I made `Run` a class because you can do something like build yourself a `PriorityAlgorithmParametersRun(Run)` class that has extra util methods that do the repeated stuff for you and whatnot, and then each of these runs here just extend that and benefit from the shared methods

Opeyem1a (changes requested):

Opeyem1a (approved):
lgtm

PR #234, opened 2023-11-16
mock student generator bug fix

### LINKED ISSUE #232: Fix mock student generator
### Issue
- The `mock_student_generator` does not assign the correct number of friends and enemies
- The enemy generation isn't safe either

justino599:
Closes #232 
Description in issue. Reasoning for fix is that since specifying 2 enemies creates two enemy relationships on each student, specifying 2 friends should also do the same. The old behaviour is that each student would be a part of a clique of 2, rather than have 2 friends

### DIFF
--- a/tests/test_benchmarking/test_data/test_simulated_data/test_mock_student_provider.py
+++ b/tests/test_benchmarking/test_data/test_simulated_data/test_mock_student_provider.py
@@ -21,6 +21,76 @@ class TestMockStudentProvider(unittest.TestCase):
         for student in students:
             self.assertIsInstance(student, Student)
 
+    def test_get__each_student_has_correct_num_friends_and_enemies(self):
+        students = MockStudentProvider(
+            MockStudentProviderSettings(
+                number_of_students=10,
+                number_of_friends=4,
+                number_of_enemies=1,
+            )
+        ).get()
+
+        for student in students:
+            num_friends = len(
+                [
+                    1
+                    for stu_id, relation in student.relationships.items()
+                    if relation == Relationship.FRIEND
+                ]
+            )
+            num_enemies = len(
+                [
+                    1
+                    for stu_id, relation in student.relationships.items()
+                    if relation == Relationship.ENEMY
+                ]
+            )
+
+            self.assertEqual(4, num_friends)
+            self.assertEqual(1, num_enemies)
+
+    def test_get__each_student_has_correct_num_friends_and_enemies_non_divisible_class_size(
+        self,
+    ):
+        students = MockStudentProvider(
+            MockStudentProviderSettings(
+                number_of_students=10,
+                number_of_friends=3,
+                number_of_enemies=1,
+            )
+        ).get()
+
+        for student in students:
+            num_friends = len(
+                [
+                    1
+                    for stu_id, relation in student.relationships.items()
+                    if relation == Relationship.FRIEND
+                ]
+            )
+            num_enemies = len(
+                [
+                    1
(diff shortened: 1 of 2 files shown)


Opeyem1a (changes requested):
I think I need to see some tests here to be convinced you fixed it. In general this function is kinda sus and was stolen from an old time. If we need it to work we should be sure it works fully yknow

tests/test_benchmarking/test_data/test_simulated_data/test_mock_student_provider.py
```diff
@@ -105,7 +175,7 @@ def test_create_mock_students__student_has_num_values_for_attribute_within_range
     def test_create_mock_students__students_have_correct_friend_and_enemy_count(self):
         for i in [1, 5, 10]:
             students = create_mock_students(
-                number_of_students=i + 1,
+                number_of_students=i * 2 + 1,
```
Opeyem1a, line 178: Q: why was this change needed?

tests/test_benchmarking/test_data/test_simulated_data/test_mock_student_provider.py
```diff
@@ -21,6 +21,76 @@ def test_get__returns_student_objects(self):
...
+            num_friends = len(
+                [
+                    1
+                    for stu_id, relation in student.relationships.items()
+                    if relation == Relationship.FRIEND
+                ]
+            )
+            num_enemies = len(
+                [
+                    1
```
Opeyem1a, line 43: nit; make this 2, I think the way it worked before could have still passed this test if you use 1

tests/test_benchmarking/test_data/test_simulated_data/test_mock_student_provider.py
```diff
@@ -21,6 +21,76 @@ def test_get__returns_student_objects(self):
         for student in students:
             self.assertIsInstance(student, Student)
 
+    def test_get__each_student_has_correct_num_friends_and_enemies(self):
```
Opeyem1a, line 24: these tests actually should go in the Helpers test class, since the function that generates students isn't in MockStudentProvider, it is just used there

tests/test_benchmarking/test_data/test_simulated_data/test_mock_student_provider.py
```diff
@@ -21,6 +21,76 @@ def test_get__returns_student_objects(self):
...
+                    1
+                    for stu_id, relation in student.relationships.items()
+                    if relation == Relationship.ENEMY
+                ]
+            )
+
+            self.assertEqual(3, num_friends)
+            self.assertEqual(1, num_enemies)
+
+    def test_get__class_size_too_small(self):
```
Opeyem1a, line 82: I'm being a bit greedy here, but you've added a few new things to care about that should be tested (I believe):
- a student's friend cannot also be their enemy, if there are enough people
- if cluster is used, we should check the cluster is a cluster

Opeyem1a (changes requested):
I'm bullying you

tests/test_benchmarking/test_data/test_simulated_data/test_mock_student_provider.py
justino599: Cause otherwise in the `i=1` case there aren't enough students in the class to satisfy the friends and enemies req

Opeyem1a: Ah gotcha

Opeyem1a (approved):
🐐

PR #236, opened 2023-11-16
Fixed bug related to student friend/enemies not being counted correctly

### LINKED ISSUE #233: Fix student calculations for friend and enemy satisfaction
### Issue
- The `student_calculations` do not correctly count the number of friends and enemies in a team

justino599:
Closes #233

### DIFF
--- a/benchmarking/evaluations/metrics/utils/student_calculations.py
+++ b/benchmarking/evaluations/metrics/utils/student_calculations.py
@@ -6,9 +6,9 @@ from api.models.student import Student
 def num_friends_satisfied(student: Student) -> int:
     count = 0
     for relation_student_id, relation in student.relationships.items():
-        if relation != Relationship.FRIEND:
+        if relation != Relationship.FRIEND or relation_student_id == student.id:
             continue
-        if relation_student_id in [t.id for t in student.team.students if t.id != student.id]:
+        if relation_student_id in [t.id for t in student.team.students]:
             count += 1
     return count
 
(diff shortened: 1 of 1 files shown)


benchmarking/evaluations/metrics/utils/student_calculations.py
```diff
@@ -9,7 +9,7 @@ def num_friends_satisfied(student: Student) -> int:
         if relation != Relationship.FRIEND:
             continue
         for teammate in student.team.students:
```
Opeyem1a, line 11: good catch here, tbh I would do something like:
```py
if relation_student_id in [t.id for t in student.team.students if t.id != student.id]:
    count += 1
```

instead of this weird loop (also rename `student_id, relation` to `relation_student_id, relation`

Opeyem1a: what do you think? @justino599 
Mostly because we actually do not want to count a friend as mattering if it's a student being friends with themselves

benchmarking/evaluations/metrics/utils/student_calculations.py
```diff
@@ -21,7 +21,7 @@ def num_enemies_satisfied(student: Student) -> int:
             continue
         count += 1
         for teammate in student.team.students:
```
Opeyem1a, line 23: similar down here it's weird to add 1 and then remove 1 if they are present imo, just only add one if they are not present

Opeyem1a:
```py
if relation_student_id not in [t.id for t in student.team.students if t.id != student.id]:
    count += 1
```

Opeyem1a (changes requested):

benchmarking/evaluations/metrics/utils/student_calculations.py
```diff
@@ -5,24 +5,21 @@
...
     count = 0
-    for student_id, relation in student.relationships.items():
+    for relation_student_id, relation in student.relationships.items():
         if relation != Relationship.ENEMY:
             continue
-        count += 1
-        for teammate in student.team.students:
-            if teammate.id == student_id:
-                count -= 1
+        if relation_student_id not in [t.id for t in student.team.students if t.id != student.id]:
```
Opeyem1a, line 21: wait I'm goofy
it should be like:
```py
if student.id == relation_student_id:
    continue
if relation_student_id not in [t.id for t in student.team.students]:
    count += 1
```

this applies for both

Opeyem1a (approved):

PR #239, opened 2023-11-17
Benchmark social algorithm with different parameters

### LINKED ISSUE #219: Benchmark social algorithm with different parameters
### Issue
- We want to see how the social algorithm behaves under different conditions
    - Bigger/smaller teams
    - One-way friendships

justino599:
### Goal
Benchmark social algorithm while varying
- Team size
- Num friends/enemies
- Student relationship type (clustered friendships, non-mutual friendships)

Closes #219

### DIFF
--- a/benchmarking/runs/social/varied_num_enemies.py
+++ b/benchmarking/runs/social/varied_num_enemies.py
@@ -51,7 +51,7 @@ class VariedNumEnemiesSocialRun(SocialRun):
                     student_provider=MockStudentProvider(student_provider_settings),
                     cache_key=f"social/varied_num_friends/{num_enemies}_enemies",
                 ),
-                algorithm_set=super().algorithms,
+                algorithm_set=SocialRun.algorithms(),
             ).run(num_runs=num_trials)
 
         if generate_graphs:
@@ -61,11 +61,11 @@ class VariedNumEnemiesSocialRun(SocialRun):
             for num_enemies in enemy_nums:
                 artifact: SimulationSetArtifact = simulation_sets[num_enemies]
                 insight_set: Dict[str, InsightOutput] = Insight.get_output_set(
-                    artifact=artifact, metrics=list(super().metrics.values())
+                    artifact=artifact, metrics=list(SocialRun.metrics().values())
                 )
 
                 average_metrics: Dict[str, Dict[str, float]] = {}
-                for metric_name in [Insight.KEY_RUNTIMES, *super().metrics.keys()]:
+                for metric_name in [Insight.KEY_RUNTIMES, *SocialRun.metrics().keys()]:
                     average_metrics[metric_name] = Insight.average_metric(
                         insight_set, metric_name
                     )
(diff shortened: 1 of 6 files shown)


benchmarking/runs/social/varied_num_enemies.py
```diff
@@ -0,0 +1,187 @@
...
+)
+from benchmarking.runs.interfaces import Run
+from benchmarking.simulation.insight import Insight, InsightOutput
+from benchmarking.simulation.simulation_set import SimulationSet, SimulationSetArtifact
+from benchmarking.simulation.simulation_settings import SimulationSettings
+
+
+class VariedNumEnemiesSocialRun(Run):
+    @staticmethod
+    def start(num_trials: int = 10, generate_graphs: bool = True):
```
Opeyem1a, line 35: we said it should default to false but we will fix these all at once when we decide on the default num trials, so this is a nit;

benchmarking/runs/social/varied_num_enemies.py
```diff
@@ -0,0 +1,187 @@
...
+        algorithms: Dict[AlgorithmType, List[AlgorithmConfig]] = {
+            AlgorithmType.RANDOM: [RandomAlgorithmConfig()],
+            AlgorithmType.WEIGHT: [WeightAlgorithmConfig()],
+            AlgorithmType.SOCIAL: [SocialAlgorithmConfig()],
+        }
+
+        # Use: simulation_sets[class_size][num_friends]: SimulationSetArtifact
+        simulation_sets: Dict[int, Dict[int, SimulationSetArtifact]] = {}
+
+        for class_size in class_sizes:
```
Opeyem1a, line 89: wait hmm why are we varying both class size and num_enemies?

benchmarking/runs/social/varied_num_friends.py
```diff
@@ -0,0 +1,189 @@
...
+
+        This run holds the team size constant at 4, then tests it with cliques that are 0.5, 1,
+        1.5, and 2 times the size of the team. This will test to see if the algorithm will be
+        able to put cliques into teams if the clique size does not match the team size.
+        The most interesting case will be the 0.5 case because it should be able to put two
+        full cliques into a single team.
+        """
+
+        clique_sizes = [2, 4, 6, 8]
+        class_sizes = [100, 200]  # list(range(50, 501, 50))
```
Opeyem1a, line 47: Oh?

benchmarking/runs/social/varied_num_friends.py
```diff
@@ -0,0 +1,189 @@
...
+
+        metrics: Dict[str, TeamSetMetric] = {
```
Opeyem1a, line 49: same comment as with seth's thing, you could extend Run into some kind of SocialRun or whatever and put all the repeated code in methods there

benchmarking/runs/social/varied_num_enemies.py
Opeyem1a: I think class size should be a separate run where only class size changes

Opeyem1a (changes requested):
comments apply to each file, hopefully makes sense

Opeyem1a (approved):
I'm not 100% convinced that num_friends.py needs to also vary the class size but we can discuss post-merge if you want

PR #251, opened 2023-11-19
Write chunk() and combine() functions

### LINKED ISSUE #250: Implement combine() and chunk() functions for parallelism
### `chunk(num_runs, num_threads)`
- Return list of length `num_threads` where each element is the number of trials that thread should do

### `combine(cache_key)`
- Search for all fragments of the given cache key and combine them into one cache, and deleting the fragments

justino599:
Closes #250

### DIFF
--- a/benchmarking/caching/utils.py
+++ b/benchmarking/caching/utils.py
@@ -40,13 +40,13 @@ def combine(cache_key: str):
         return
 
     # Get initial list of files where the cache should be
-    fragments = [
-        f for f in os.listdir(cache_location) if re.match(r"fragment_\d+\.json", f)
-    ]
-    # Remove .json
-    fragments = [f[:-5] for f in fragments]
+    fragments = []
+    for f in os.listdir(cache_location):
+        match = re.match(r"(fragment_\d+)\.json", f)
+        if match:
+            fragments.append(match.group(1))
 
-    info: List[Dict[str, any]] = []
+    fragment_data: List[Dict[str, any]] = []
 
     for fragment in fragments:
         cache = SimulationCache(cache_key + "/" + fragment)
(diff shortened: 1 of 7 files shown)


benchmarking/caching/utils.py
```diff
@@ -0,0 +1,99 @@
...
+    # Write out resulting combined cache to cache_key
+    full_cache = SimulationCache(cache_key)
+    full_cache.clear()
+    full_cache.save(**result_cache)
+
+    # Remove fragments
+    shutil.rmtree(cache_location)
+
+
+if __name__ == "__main__":
```
Opeyem1a, line 98: remove

benchmarking/caching/utils.py
```diff
@@ -0,0 +1,99 @@
...
+    # If cache not found, return
+    if not path.exists(cache_location):
+        return
+
+    # Get initial list of files where the cache should be
+    fragments = [
+        f for f in os.listdir(cache_location) if re.match(r"fragment_\d+\.json", f)
+    ]
+    # Remove .json
+    fragments = [f[:-5] for f in fragments]
```
Opeyem1a, line 47: 😬

benchmarking/caching/utils.py
```diff
@@ -0,0 +1,99 @@
...
+
+    info: List[Dict[str, any]] = []
```
Opeyem1a, line 49: nit;

is info a good variable name here

benchmarking/caching/utils.py
```diff
@@ -0,0 +1,99 @@
...
+    # Sort by timestamp so that the newest fragment has metadata priority
+    sorted_info = sorted(info, key=lambda x: x["metadata"]["timestamp"])
+
+    result_cache: Dict[str, any] = {
+        "metadata": {},
+        "team_sets": [],
+        "runtimes": [],
+    }
+    for fragment_data in sorted_info:
+        result_cache["metadata"].update(fragment_data["metadata"])
```
Opeyem1a, line 85: use `fragment_data.get("metadata")` instead

applies for the next 2 lines too

Opeyem1a (changes requested):
minor changes

benchmarking/caching/utils.py
justino599: Put a check earlier in the file instead because if this doesn't exist, the fragment is corrupted or something and should be skipped

benchmarking/caching/utils.py
```diff
@@ -0,0 +1,111 @@
...
+            print(
+                f"[WARNING]: Cache fragment missing critical data, skipping {fragment}",
+                file=sys.stderr,
+            )
+            continue
+
+        # Commit hashes must match
+        if (
+            len(fragment_data) > 0
+            and metadata["commit_hash"] != fragment_data[0]["metadata"]["commit_hash"]
```
Opeyem1a, line 75: On second thought, why only check at index 0 here? I'm not sure I understand the data structure but this check and the error message it throws don't seem to match

Opeyem1a: Oh nevermind I think I get it, feels a little gross of a way to write it though, probably just a nit; but idk something about this logic feels off. 

I think it's the fact we assume the first fragment is the source of truth? Idk

Opeyem1a (approved):

PR #257, opened 2023-11-21
Use serializers in api data loader

### LINKED ISSUE #180: Use serializers in API loader
serializers for models have been created and should be used on the django API side when possible

ketphan02:
Closes #180


Opeyem1a (approved):
Cool 💪

PR #265, opened 2023-11-23
Fix calculation of max_team_size

### LINKED ISSUE #264: Social alg generates teams larger than they should be
### Issue
Social alg is making teams of size 6 when it should be 5

### Reproduce
- Provide 50 students clustered into cliques of 6
- Tell social algorithm to make 10 teams
- Will create teams with sizes `6, 6, 6, 6, 6, 6, 6, 6, 2, 0`

justino599:
## Problem
The social algorithm finds cliques of students up to a clique size of `max_team_size`. The problem is that `max_team_size` is not being calculated correctly. It assumed that `num_students` is not evenly divisible by `num_teams`. The result is that `max_team_size` is one too big in the case that it is evenly divisible and social was finding cliques that were too big.

Closes #264


Opeyem1a (approved):
Oh good on my part, good catch

PR #267, opened 2023-11-23
Use the correct number of friends for social runs

### LINKED ISSUE #266: Social runs not using correct number of friends

justino599:
I done goofed this one, silly mistake. It was intended for the clique size to be equal to the team size, which is 5 for these two runs

Closes #266


Opeyem1a (approved):
Lgtm

PR #275, opened 2023-11-28
fix tests

Opeyem1a:
The problem is that the cache returns results in order of timestamp, but the output from the simulation run that creates the cache returns in order of fragment id.

Depending on how long it takes processes to spin up, there is no guarantee that the timestamp order matches the fragment id order, thus creating a flaky test.

### DIFF
--- a/api/models/team_set/model.py
+++ b/api/models/team_set/model.py
@@ -6,7 +6,7 @@ from api.models.team import Team
 
 @dataclass
 class TeamSet:
-    _id: int = None
+    _id: str = None
     name: str = None
     teams: List[Team] = field(default_factory=list)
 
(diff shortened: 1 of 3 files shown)


Opeyem1a:
* main
  * **PR #275** <a href="https://app.graphite.dev/github/pr/Teamable-Analytics/algorithms/275?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/Teamable-Analytics/algorithms/275?utm_source=stack-comment).

justino599 (approved):
Ah true true, I overlooked that detail when checking over these tests, good catch

PR #276, opened 2023-11-28
Dockerize api for prod deployment

ketphan02:


ketphan02:
Currently in the process of deploying to UBC server.

Dockerfile.dev
```diff
@@ -0,0 +1,27 @@
...
+ENV PYTHONUNBUFFERED 1
+
+# Upgrading pip version
+RUN python3 -m pip install --upgrade pip
+
+# Copying requirement file
+COPY requirements.txt .
+
+## Installing dependencies
+RUN python3 -m pip install gunicorn
```
keyvankhademi, line 21: Is there a reason you didn't put gunicorn in the requirements.txt?

nginx/nginx.conf
```diff
@@ -0,0 +1,18 @@
+upstream django_app {
```
keyvankhademi, line 1: I would prefer to have the config in the gamification style
There is no need to use upstream here
it's like creating a function with a single like that call another function.

docker-compose.prod.yml
```diff
@@ -0,0 +1,21 @@
+services:
+  nginx:
+    build:
+      context: nginx
+      dockerfile: Dockerfile
```
keyvankhademi, line 5: i don't understand the Dockerfile. Couldn't we simply use a pre-built nginx container?

keyvankhademi (approved):

keyvankhademi:
We don't need docker for dev. We should not develop in docker. You can remove the dev ones.

Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+ENV PYTHONUNBUFFERED 1
+
+# install system dependencies
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends gcc
+
+# lint
+RUN python -m pip install --upgrade pip
+COPY . /usr/src/app/
+RUN python -m pip install black==23.7.0
```
keyvankhademi, line 16: You don't need to run lint in docker. It's just a waste of time. You can run it in the pipeline if you want to keep the code clean.

Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+
+
+# install python dependencies
+COPY ./requirements.txt .
+RUN pip wheel --no-cache-dir --no-deps --wheel-dir /usr/src/app/wheels -r requirements.txt
+
+
+FROM python:3.9-slim-buster
```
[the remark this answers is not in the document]
keyvankhademi: Can you not have all of it from 1 image?

Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+
+# create directory for the app user
+RUN mkdir -p /home/app
+
+# create the app user
+RUN addgroup --system app && adduser --system --group app
```
keyvankhademi, line 30: I wouldn't deal with the user in the container unless it's absolutely necessary. Just let the root handle everything.

Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+ENV APP_HOME=/home/app/web
+RUN mkdir $APP_HOME
+WORKDIR $APP_HOME
+
+# install dependencies
+RUN apt-get update && apt-get install -y --no-install-recommends netcat
+COPY --from=builder /usr/src/app/wheels /wheels
+COPY --from=builder /usr/src/app/requirements.txt .
+RUN pip install --upgrade pip
+RUN pip install --no-cache /wheels/*
```
keyvankhademi, line 43: See now you are installing everything twice. If you use just 1 image and handle everything there it will be more simple and efficient.

Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+
+# copy project
+COPY . $APP_HOME
+
+# chown all the files to the app user
+RUN chown -R app:app $APP_HOME
+
+# change to the app user
+USER app
```
keyvankhademi, line 52: If you just let the root user handle everything you don't need these lines either.

keyvankhademi (approved):

Opeyem1a (approved):

PR #277, opened 2023-11-29
Fix weird algorithm to use different utility function

ketphan02:
### What in this PR?
- 2 main issues:
  -  The algorithms from the papers have some assumptions. These assumptions require a list of all students, therefore we cannot put them in the constructor, and if we put them in the `generate` function, it affects the runtime.
  - The algorithms are not well-written. They can only hold 1 type of utility. Redo them so they can use different utility functions, depending on which we pass in
- Minor issues: remove all `projects` and replace them with `team`. Project requirements will be moved to team requirements


api/ai/algorithm_runner.py
```diff
@@ -52,12 +54,25 @@ def __init__(
...
             team_generation_options=self.team_generation_options,
             algorithm_options=self.algorithm_options,
             algorithm_config=self.algorithm_config,
         )
-        return algorithm.generate(students)
+        self.algorithm = algorithm
+
+        try:
+            algorithm.prepare(students)
+        except NotImplementedError:
```
Opeyem1a, line 73: Idk if I like this, just have the prepare method on the abstract class `pass` instead of throwing not implemented. It's not something we need every algorithm to have, and it is acceptable for the ones that don't need to just do nothing at that step

api/ai/algorithm_runner.py
```diff
@@ -52,12 +54,25 @@ def __init__(
     def generate(self, students: List[Student]) -> TeamSet:
         # the algorithm classes internally track generated teams, so a new instance of the
         #   algorithm class MUST be created to run a new generation without side effects
+        # This generated instance is initialized in the prepare method
+        return self.algorithm.generate(students)
+
+    def prepare(self, students: List[Student]) -> None:
```
Opeyem1a, line 60: Since I don't think `prepare` should be excluded from runtime, I think there's no need for `AlgorithmRunner.prepare`. 

Instead, I propose that `AlgorithmRunner.generate` just calls `algorithm.prepare` and then `algorithm.generate`

api/ai/interfaces/algorithm.py
```diff
@@ -29,6 +29,12 @@ def __init__(
     def generate(self, students: List[Student]) -> TeamSet:
         raise NotImplementedError
 
+    def prepare(self, students: List[Student]) -> None:
+        """
+        This part is used to set up the algorithm before running the generation without affecting run time
+        """
+        raise NotImplementedError
```
Opeyem1a, line 36: `pass`

api/ai/interfaces/algorithm_options.py
```diff
@@ -247,13 +249,10 @@ def get_schema() -> Schema:
 
 @dataclass
 class MultipleRoundRobinAlgorithmOptions(AlgorithmOptions):
-    projects: List[Project]
+    utility_function: Callable[[Student, TeamShell], float]
```
Opeyem1a, line 252: Utility functions seem like config things to me, unless it's intended that they change per call of the algorithm?

api/ai/multiple_round_robin_with_adjusted_winner_algorithm/utils.py
```diff
@@ -0,0 +1,33 @@
...
+
+from api.models.student import Student
+from api.models.team import TeamShell, Team
+
+
+def is_ordered_envy_freeness_up_to_one_item(
+    team_i: Team, team_j: Team, utility_function: Callable[[Student, TeamShell], float]
+) -> bool:
+    """
+    True if value of team_i is less than value of team_j and if we remove 1 student from team_j, the value of team_i is
```
Opeyem1a, line 11: I don't need this comment to be an essay, but it feels like it does not describe the thing this function actually does, notably the shenanigans around requirements

benchmarking/simulation/simulation.py
```diff
@@ -156,6 +156,7 @@ def run_trial_batch(
             else:
                 students = settings.student_provider.get()
 
+            runner.prepare(students)
```
Opeyem1a, line 159: remove

Opeyem1a (changes requested):

api/ai/interfaces/algorithm_options.py
ketphan02: Sounds right, it's a config

api/ai/interfaces/algorithm.py
```diff
@@ -29,6 +29,12 @@ def __init__(
     def generate(self, students: List[Student]) -> TeamSet:
         raise NotImplementedError
 
+    def prepare(self, students: List[Student]) -> None:
+        """
+        This part is used to set up the algorithm before running the generation without affecting run time
```
Opeyem1a, line 34: comment is now incorrect

Opeyem1a (approved):
Trusting you to fix the thing I commented before merging

PR #278, opened 2023-11-29
Generate Data for Priority Algorithm Diversify Gender Min 2 Benchmarks

Hedgemon4:
### Description
- generated data for the diversify gender min 2 scenario


Hedgemon4:
This data is outdated

PR #284, opened 2023-11-30
Diversify Gender Priority Algorithm Report

Hedgemon4:
### Description
- write up about the graph and output from the priority algorithm diversify gender min 2 runs


Hedgemon4:
This uses outdated data and is not consistent with how we are now investigating the algorithm parameters.

PR #289, opened 2023-12-05
Fix API output data is string

ketphan02:
Our API currently return string instead of json, this reduces readability

Before
image
After
image


Opeyem1a (commented):
Wait isn't it already returning a dictionary from the TeamSetSerializer? If not, why not?

I'm not understanding why the original problem exists to begin with?

api/api/views/generate_teams.py
```diff
@@ -42,6 +44,5 @@ def generate_teams(self, request):
 
         serialized_team_set = TeamSetSerializer().encode(team_set)
```
Opeyem1a, line 45: I think we forgot to call `.default` instead of `.encode`, I think this was an artifact of us changing the interface of serializers a while ago.

Opeyem1a (changes requested):

Opeyem1a (approved):

PR #294, opened 2023-12-10
Use the same set of students within same trial

### LINKED ISSUE #286: Use the same set of students within same trial

justino599:
Closes #286

### DIFF
--- a/benchmarking/simulation/simulation.py
+++ b/benchmarking/simulation/simulation.py
@@ -1,5 +1,5 @@
 import os
-import random
+import re
 import time
 import uuid
 from multiprocessing import Pool
@@ -38,7 +38,10 @@ class Simulation:
         self.team_sets = []
         self.run_times = []
 
-    def run(self, num_runs: int) -> SimulationArtifact:
+    def run(self, num_runs: int, seeds: List[int]) -> SimulationArtifact:
+        if len(seeds) != num_runs:
+            raise ValueError("Must provide a seed for each run")
+
         if self.settings.cache_key:
             cache = SimulationCache(self.settings.cache_key)
             if cache.exists():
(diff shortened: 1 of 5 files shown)


benchmarking/simulation/simulation.py
```diff
@@ -50,6 +53,13 @@ def run(self, num_runs: int) -> SimulationArtifact:
                 else:
                     num_runs -= len(cached_team_sets)
 
+        completed_run_indexes = []
+        for team_set in self.team_sets:
+            team_set_id = team_set._id
+            match = re.search(r"run_index_(\d+)", team_set_id)
```
Opeyem1a, line 59: we need to be like pretty confident we don't break this id naming convention

Opeyem1a: I'd feel more comfortable if this was a number just placed somewhere.

I was envisioning more like we save the used run indexes in the metadata of the cache file

And then merge those back together in the combine function

And then we can just get the list of used run indexes all at once

benchmarking/simulation/simulation.py
```diff
@@ -104,9 +129,10 @@ def run(self, num_runs: int) -> SimulationArtifact:
 
 def run_trial_batch(
     fragment: int,
-    num_runs_for_batch: int,
+    run_indexes_for_batch: List[int],
```
Opeyem1a, line 132: I don't think this should get to this level. I think this function should only have awareness of the num runs and the seeds it needs

We should have already only passed the seeds at the needed indexes to this function, essentiallyh

Opeyem1a (changes requested):

Opeyem1a (approved):
Prolly make a ticket to add tests (like actually make a ticket, it's low prio to do now)

PR #304, opened 2023-12-19
Add fixed attribute generation

ketphan02:


benchmarking/data/simulated_data/mock_student_provider.py
```diff
@@ -241,3 +266,30 @@ def attribute_values_from_range(
         replace=False,
         generator=_generator,
     )
+
+
+def _generate_attribute_values_without_probability(
+    attribute_ranges: Dict[int, AttributeRangeConfig], class_size: int
+):
```
Opeyem1a, line 273: nit;

return type

tests/test_benchmarking/test_data/test_simulated_data/test_mock_student_provider.py
```diff
@@ -344,21 +344,29 @@ def test_attribute_values_from_range__multiple_values_for_attribute_are_selected
             # FIXME: This test is (technically) a little flaky 😬
             # run this multiple times so that this test cannot accidentally
             #   pass even if the values are selected with replacement
-            values_1 = attribute_values_from_range([10, 20, 30], num_values=3)
+            values_1 = attribute_values_from_range(
```
Opeyem1a, line 347: I think you definitely gotta add more test cases to be sure this new addition works

benchmarking/data/simulated_data/mock_student_provider.py
```diff
@@ -241,3 +266,30 @@ def attribute_values_from_range(
...
+    probabilistic_attributes = {}
+    for attribute_id, attribute_range_config in attribute_ranges.items():
+        if isinstance(attribute_range_config[0], (int, AttributeValueEnum)):
+            continue
+
+        # Assume if the range is probabilistic, the number of value per attribute is 0
+        if (sum([_[1] for _ in attribute_range_config])) != 1:
+            raise ValueError(f"Probabilistic attribute range must sum to 100$")
```
Opeyem1a, line 281: 100%*

benchmarking/data/simulated_data/mock_student_provider.py
```diff
@@ -241,3 +266,30 @@ def attribute_values_from_range(
...
+
+        probabilistic_attributes[attribute_id] = []
+        for attribute_range in attribute_range_config:
+            probabilistic_attributes[attribute_id].extend(
+                [
+                    attribute_range[0]
+                    if isinstance(attribute_range[0], int)
+                    else attribute_range[0].value
+                    for _ in range(int(attribute_range[1] * class_size))
```
Opeyem1a, line 290: Does this not break when because `int()` will floor the value by definition, so it's possible that this array ends up with less items in it than `class_size`

benchmarking/data/simulated_data/mock_student_provider.py
```diff
@@ -241,3 +266,30 @@ def attribute_values_from_range(
...
+
+        # Assume if the range is probabilistic, the number of value per attribute is 0
+        if (sum([_[1] for _ in attribute_range_config])) != 1:
+            raise ValueError(f"Probabilistic attribute range must sum to 100$")
+
+        probabilistic_attributes[attribute_id] = []
+        for attribute_range in attribute_range_config:
+            probabilistic_attributes[attribute_id].extend(
+                [
+                    attribute_range[0]
```
Opeyem1a, line 287: nit;

Feel like we should just name this variable, like `attribute_value, attribute_probability = attribute_range`

benchmarking/data/simulated_data/mock_student_provider.py
```diff
@@ -241,3 +266,30 @@ def attribute_values_from_range(
...
+
+def _generate_attribute_values_without_probability(
+    attribute_ranges: Dict[int, AttributeRangeConfig], class_size: int
+):
+    probabilistic_attributes = {}
+    for attribute_id, attribute_range_config in attribute_ranges.items():
+        if isinstance(attribute_range_config[0], (int, AttributeValueEnum)):
+            continue
+
+        # Assume if the range is probabilistic, the number of value per attribute is 0
```
Opeyem1a, line 279: wait what does this mean and why?

benchmarking/data/simulated_data/mock_student_provider.py
```diff
@@ -241,3 +266,30 @@ def attribute_values_from_range(
...
         generator=_generator,
     )
+
+
+def _generate_attribute_values_without_probability(
+    attribute_ranges: Dict[int, AttributeRangeConfig], class_size: int
+):
+    probabilistic_attributes = {}
+    for attribute_id, attribute_range_config in attribute_ranges.items():
+        if isinstance(attribute_range_config[0], (int, AttributeValueEnum)):
```
Opeyem1a, line 276: Wait so what happens if this is the case then? Like the dictionary would just have nothing in it for these attributes

Opeyem1a (changes requested):

ketphan02:
Fixed in #314

PR #310, opened 2023-12-21
Add Option For Exponent Base in Priority Satisfaction Metric

Hedgemon4:
### Description
- added an option for the exponent base in the priority satisfaction metric

### DIFF
--- a/benchmarking/evaluations/metrics/priority_satisfaction.py
+++ b/benchmarking/evaluations/metrics/priority_satisfaction.py
@@ -11,10 +11,18 @@ class PrioritySatisfaction(TeamSetMetric):
     Calculates the number of priorities satisfied by the algorithm
     """
 
-    def __init__(self, priorities: List[Priority], is_linear: bool, *args, **kwargs):
+    def __init__(
+        self,
+        priorities: List[Priority],
+        is_linear: bool,
+        exponent_base: int = 2,
+        *args,
+        **kwargs
+    ):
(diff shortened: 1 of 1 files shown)


benchmarking/evaluations/metrics/priority_satisfaction.py
```diff
@@ -11,10 +11,18 @@ class PrioritySatisfaction(TeamSetMetric):
     Calculates the number of priorities satisfied by the algorithm
     """
 
-    def __init__(self, priorities: List[Priority], is_linear: bool, *args, **kwargs):
+    def __init__(
+        self,
+        priorities: List[Priority],
+        is_linear: bool,
+        exponent_base: int = 2,
```
Opeyem1a, line 18: nit;

uhhhh, how does this affect existing code? Feel like it should default to whatever the actual prio algorithm is using, like import NUM_BRACKETS

Opeyem1a (approved):

benchmarking/evaluations/metrics/priority_satisfaction.py
Hedgemon4: This is what it was before. It was not the same as num buckets previously.

Hedgemon4: Let me know if this is fine

Opeyem1a: O

Opeyem1a (approved):

PR #312, opened 2023-12-23
308 build function to query teams for number of females

Hedgemon4:


Hedgemon4:
This was handled in #313

PR #313, opened 2023-12-23
Create Function to Count the Number of Students with a Certain Attribute

Hedgemon4:
### Description
- counts the number of students on each team who have a certain attribute
- returns a list of lists where the outer list has length `num_trials` and the inner list length is the number of teams in the `team_set` file


Opeyem1a (approved):
Skimmed

PR #315, opened 2023-12-26
Add API key validation

ketphan02:


api/api/views/generate_teams.py
```diff
@@ -28,6 +29,15 @@ def generate_teams(self, request):
         6. Return JSON to a happy user :) 🚀!
         """
         try:
+            request_headers = dict(request.headers)
```
Opeyem1a, line 32: I feel like this should be middleware, not just in this view

utils/api_auth.py
```diff
@@ -0,0 +1,37 @@
+import os
+from bcrypt import hashpw, gensalt, checkpw
+
+from pathlib import Path
+import dotenv
+
+env_file = next((Path(__file__).parent.parent / "env").glob("*.env"))
```
Opeyem1a, line 7: what does this do?

utils/api_auth.py
```diff
@@ -0,0 +1,37 @@
...
+
+    This function only serves to encrypt a new API token for customers and should never get called in production.
+    """
+    return hashpw(
+        password=api_token.encode("utf-8"),
+        salt=gensalt(),
+    )
+
+
+def check_hash(api_token, encrypted_api_token):
```
Opeyem1a, line 23: nit; 

does this return bool? generally why no types

api/api/settings/__init__.py
```diff
@@ -29,7 +25,7 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = True
 
-ALLOWED_HOSTS = list(os.environ.get("DJANGO_ALLOWED_HOSTS", "").split(" "))
+ALLOWED_HOSTS = ["*"]
```
Opeyem1a, line 28: why?

Opeyem1a (changes requested):

api/api/settings/__init__.py
ketphan02: This is dev, we still have this on prod. This makes the API work locally even without an `.env`

utils/api_auth.py
ketphan02: This loads either `.env` or `production.env`  inside of the `env` folder (I adopted this structure from Gamification)

Opeyem1a (approved):

PR #316, opened 2023-12-26
Fix API production refuses to handle localhost

ketphan02:


api/api/settings/prod.py
```diff
@@ -94,7 +94,7 @@
 ]
 
 # CORS
-CORS_ALLOWED_ORIGINS = ["*"]
+CORS_ALLOW_ALL_ORIGINS = True
```
Opeyem1a, line 97: On prod?

Opeyem1a (changes requested):

requirements.txt
```diff
@@ -10,6 +10,7 @@ async-lru==2.0.4
 attrs==23.1.0
 Babel==2.13.1
 backcall==0.2.0
+bcrypt==4.1.2
```
Opeyem1a, line 13: also why?

Opeyem1a (changes requested):

api/api/settings/prod.py
ketphan02: Yes, since this is public API

requirements.txt
ketphan02: I save the encrypted key as a password, therefore, even when the key is leaked (not likely), it is less likely to exploit our service

Opeyem1a (approved):

PR #317, opened 2023-12-26
Add documentation page

ketphan02:


api/api/templates/docs.html
```diff
@@ -0,0 +1,171 @@
+{% load static %}
+<!DOCTYPE html>
```
Opeyem1a, line 2: I thought this page was going to be on the other repo? There is no explicit plan for this repo to be deployed with any kind of pretty URL, so I'm not sure it makes sense for this to be here.

I also really really dislike putting bootstrap, css, and html into this repo that is supposed to be a microservice purely for algorithm logic

Opeyem1a (changes requested):

api/api/templates/docs.html
ketphan02: I see, should we then close this PR and transfer this into another project?

Opeyem1a: Yes

Opeyem1a: Yes

ketphan02:
Not suit this PR, move to different one

PR #320, opened 2024-01-13
Create simple diversity runs

### LINKED ISSUE #319: Make simple runs with diversification goals
We want to create some simple scenarios with a low number of students where we can easily compute the theoretical maximum score.

- We want to use 12 students for each of the cases
- These students need to be mocked, not generated. We want the same students each time
- Still need multiple trials to account for randomness within the algorithm
- We need runs with 1, 2, and 3 diversity constraints
	- Diversify gender min 2
	- Diversify GPA min 2
	- Diversify age min 2

justino599:
Ended up doing 5 constraints for two reasons:
1. It makes the scenario harder
2. It makes it easier to generate a class with unique attributes to ensure a non-trivial solution to the scenario

Closes #319


justino599:
Was not used for the paper

PR #325, opened 2024-01-20
Timeslot coverage metrics

ketphan02:


benchmarking/evaluations/metrics/average_timeslot_coverage.py
```diff
@@ -0,0 +1,58 @@
...
+
+class AverageTimeslotCoverage(TeamSetMetric):
+    """
+    Calculate the average mutual timeslots between students in a team and average over all teams.
+    """
+
+    def __init__(
+        self,
+        available_timeslots: List[int],
+        timeslot_attribute_value: int = ScenarioAttribute.TIMESLOT_AVAILABILITY.value,
```
Opeyem1a, line 16: kind of a nit;, but it's super confusing to call this a value when it is, by definition, an id

Opeyem1a (changes requested):

Opeyem1a (approved):

PR #326, opened 2024-01-20
Fix MockStudentProvider validation

ketphan02:
- `self.attribute_ranges` has the type `Dict[int, AttributeRangeConfig]` but this is checking `self.attribute_ranges[0]`, which might not exist. We can move this down inside of the loop.
- `total_chance` is a float, so should compare with margin of error


Opeyem1a (approved):
Not sure how you found this or if it was needed, but I trust the test suite so good to go

PR #333, opened 2024-01-24
Write runtime test for priority

### LINKED ISSUE #332: Runtimes of of default parameters

justino599:
Tests the runtime of the priority algorithm for the January paper. Simulation cache data located at https://github.com/Teamable-Analytics/algorithms-simulation-cache/pull/1

Closes #332


benchmarking/runs/priority_algorithm/default_parameters/runtimes.py
```diff
@@ -0,0 +1,148 @@
...
+                    value=self.value_of_female,
+                ),
+            ),
+            DiversityGoal(
+                DiversifyType.DIVERSIFY,
+                ScenarioAttribute.GPA.value,
+                tokenization_constraint=TokenizationConstraint(
+                    direction=TokenizationConstraintDirection.MIN_OF,
+                    threshold=2,
+                    value=self.value_of_female,
```
Opeyem1a, line 141:
```suggestion
                    value=self.value_of_a,
```

Opeyem1a (changes requested):

benchmarking/runs/priority_algorithm/default_parameters/runtimes.py
```diff
@@ -0,0 +1,150 @@
...
+                    )
+                    print(
+                        f"Runtime: {Insight.average_metric(insight_output, Insight.KEY_RUNTIMES)}"
+                    )
+                    for name in metrics.keys():
+                        print(f"{name}: {Insight.average_metric(insight_output, name)}")
+                    print()
+
+
+class OneDiversityScenario(Scenario):
```
Opeyem1a, line 126: We should just decide now if we're cool with defining scenario in the runs like this

And/or what the pattern is. Is it like only reusable scenarios go in the scenario file and the runs files can contain all the specifically made scenarios?

benchmarking/runs/priority_algorithm/default_parameters/runtimes.py
```diff
@@ -0,0 +1,150 @@
...
+                            False,
+                        ),
+                        "AverageSocialSatisfaction": AverageSocialSatisfaction(
+                            metric_function=is_happy_team_1hp_friend,
+                        ),
+                    }
+                    insight_output = Insight.get_output_set(
+                        artifact, list(metrics.values())
+                    )
+                    print(
```
Opeyem1a, line 115: Spooky prints, did we start doing this in runs generally?

Opeyem1a (changes requested):

benchmarking/runs/priority_algorithm/default_parameters/runtimes.py
justino599: This run only produced runtimes, no graphs, so printing is the simplest way of getting the data here

benchmarking/runs/priority_algorithm/default_parameters/runtimes.py
justino599: My logic for the scenario in the run file is that we end up creating so many, one for each run, that they are better of kept in the same file. The idea of the scenarios folder was to reuse common scenarios, but most of the time we were running very specific scenarios and I felt it would make a mess of the scenarios folder. This PR obviously isn't bad, but branch 321 has a ton more, so that's where I started that pattern.

An issue #381 has been created to address the docs change needed for this

Opeyem1a (approved):

PR #335, opened 2024-01-24
Fix Priority Algorithm's internal scoring function for concentration

ketphan02:
Fix bugs from #334 

From Discord message:
> I got `{'PrioritySatisfaction': {'AlgorithmType.PRIORITY-PriorityAlgorithm': 1.0833333333333333}}`

This is caused by always defaulting max_num_choices to 1.


Opeyem1a (approved):
Oh right good catch

PR #343, opened 2024-01-27
Add solo status metric

ketphan02:


benchmarking/evaluations/metrics/average_solo_status.py
```diff
@@ -0,0 +1,38 @@
...
+                        for attr in student.attributes.get(attribute_id)
+                    ]
+                )
+
+                for attribute_group in attribute_groups:
+                    if (
+                        attribute_group in all_attribute_values_in_team
+                        and all_attribute_values_in_team.get(attribute_group) == 1
+                    ):
+                        is_solo = True
```
Opeyem1a, line 34: maybe I just don't get the logic but this is oddly confusing code to follow for how long it is

Opeyem1a (changes requested):

benchmarking/evaluations/metrics/average_solo_status.py
```diff
@@ -0,0 +1,38 @@
...
+
+        if len(minority_groups) == 0:
+            raise ValueError("minority_groups cannot be empty")
+
+    def calculate(self, team_set: TeamSet) -> float:
+        total = 0.0
+        for team in team_set.teams:
+            is_solo = False
+            for attribute_id, attribute_groups in self.minority_groups.items():
+                all_attribute_values_in_team = Counter(
```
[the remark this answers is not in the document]
ketphan02: Transform the attribute into a dictionary of { attribute_value: num_appear }

benchmarking/evaluations/metrics/average_solo_status.py
ketphan02: If the value that we want to count solo_status has the counter of 1, we says the team is solo-status-ed

benchmarking/evaluations/metrics/average_solo_status.py
```diff
@@ -0,0 +1,44 @@
...
+from api.models.team_set import TeamSet
+from benchmarking.evaluations.interfaces import TeamSetMetric
+
+
+class AverageSoloStatus(TeamSetMetric):
+    """
+    Calculate the average number of students having solo status in a team
+    """
+
+    def __init__(self, minority_groups: Dict[int, List[int]], *args, **kwargs):
```
Opeyem1a, line 13: nit;

I think what made/makes this confusing is the lack of context/explanation on the meaning of `minority groups`

benchmarking/evaluations/metrics/average_solo_status.py
```diff
@@ -0,0 +1,44 @@
...
+        total_students = 0.0
+        for team in team_set.teams:
+            for student in team.students:
+                total_students += 1
+                for attribute_id, attribute_groups in self.minority_groups.items():
+                    # Create a dictionary of attribute_values and number of appearance in the team
+                    all_attribute_values_in_team = Counter(
+                        [
+                            attr
+                            for student in team.students
```
Opeyem1a, line 31: this is confusing because the variable name `student` exists in the scope of the list comprehension and also in the scope of the second layer for loop - hard to tell which this one means and life gets complicated

benchmarking/evaluations/metrics/average_solo_status.py
```diff
@@ -0,0 +1,44 @@
...
+                        ]
+                    )
+
+                    for attr in student.attributes.get(attribute_id):
+                        if (
+                            attr in self.minority_groups.get(attribute_id)
+                            and all_attribute_values_in_team.get(attr) == 1
+                        ):
+                            total_solo_status_students.add(student.id)
+                            break
```
Opeyem1a, line 42: An important comment to leave somewhere is that a student can only receive solo status once

benchmarking/evaluations/metrics/average_solo_status.py
```diff
@@ -0,0 +1,44 @@
+from collections import Counter
+from typing import Dict, List
+
+from api.models.team_set import TeamSet
+from benchmarking.evaluations.interfaces import TeamSetMetric
+
+
+class AverageSoloStatus(TeamSetMetric):
+    """
+    Calculate the average number of students having solo status in a team
```
Opeyem1a, line 10: Like this comment should explain a fair deal more about what is happening here

Opeyem1a (changes requested):
Okay I get it now and I mostly trust it because of your tests, just some small comments because this is honestly kinda mega confusing and its written without the comments/context it needs.

This is an example of something kinda nonintuitive that should have a little more lore dropped in. Address the non-nit comments and we're gucci

benchmarking/evaluations/metrics/average_solo_status.py
Opeyem1a: you didn't resolve this (?)

benchmarking/evaluations/metrics/average_solo_status.py
```diff
@@ -0,0 +1,51 @@
...
+    def __init__(self, minority_groups_map: Dict[int, List[int]], *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.minority_groups_map = minority_groups_map
+
+        if len(minority_groups_map) == 0:
+            raise ValueError("minority_groups cannot be empty")
+
+    def calculate(self, team_set: TeamSet) -> float:
+        total_solo_status_students = set()
+        total_students = 0.0
```
Opeyem1a, line 29: nit;

is this supposed to be a count of all unique students?

if so, it is non-intuitive why it's defaulted to the float of 0, instead of just 0 (I know why, but still, burned an extra few minutes trying to understand this one choice)

benchmarking/evaluations/metrics/average_solo_status.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def __init__(self, minority_groups_map: Dict[int, List[int]], *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.minority_groups_map = minority_groups_map
+
+        if len(minority_groups_map) == 0:
+            raise ValueError("minority_groups cannot be empty")
+
+    def calculate(self, team_set: TeamSet) -> float:
+        total_solo_status_students = set()
```
Opeyem1a, line 28: nit;

naming this `total_solo_status_students` implies it is a count, when it isn't,a nd then variable below is named similarly, even though they store completely different types of data

Opeyem1a (changes requested):

Opeyem1a (approved):

PR #356, opened 2024-01-31
Add stdev of social satisfaction among teams

### LINKED ISSUE #355: Add getting stdev of social metrics

justino599:
Simple addition to help get the stdev of social satisfaction among teams

Closes #355

### DIFF
--- a/benchmarking/evaluations/metrics/average_social_satisfied.py
+++ b/benchmarking/evaluations/metrics/average_social_satisfied.py
@@ -18,5 +18,5 @@ class AverageSocialSatisfaction(TeamSetMetric):
             / team_set.num_teams
         )
 
-    def stdev(self, team_set: TeamSet) -> float:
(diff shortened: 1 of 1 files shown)


benchmarking/evaluations/metrics/average_social_satisfied.py
```diff
@@ -16,3 +17,6 @@ def calculate(self, team_set: TeamSet) -> float:
             sum([self.metric_function(team) for team in team_set.teams])
             / team_set.num_teams
         )
+
+    def stdev(self, team_set: TeamSet) -> float:
```
Opeyem1a, line 21: nit;

we kinda have been calling this `calculate_stdev` in other metric classes

tbh we should probably just add this to the abstract class since we're using it on so many metrics, I don't want it to be called like 7 things from class to class (this part can be a separate issue/PR)

Opeyem1a (approved):

PR #370, opened 2024-02-25
Metrics API

ketphan02:


api/api/urls/__init__.py
```diff
@@ -19,11 +19,12 @@
 
 from rest_framework.routers import DefaultRouter
 
-from api.api.views import GenerateTeamsViewSet
+from api.api.views import GenerateTeamsViewSet, EvaluateTeamSetViewSet
 
 router = DefaultRouter()
 
 router.register(r"generate", GenerateTeamsViewSet, basename="generate_teams")
+router.register(r"evaluate", EvaluateTeamSetViewSet, basename="evaluate_team_set")
```
Opeyem1a, line 27: why is the basename not just `evaluate`

api/api/validators/evaluate_team_set_validator.py
```diff
@@ -0,0 +1,199 @@
...
+
+        if "id" in team_set:
+            Schema(str).validate(team_set["id"])
+            # Check if team_set[id] is a number
+            try:
+                Schema(int).validate(int(team_set["id"]))
+            except ValueError:
+                raise ValueError("team_set[id] must be a number")
+
+            Schema(str).validate(team_set["name"])
```
Opeyem1a, line 35: nit;

you don't actually need a name to calculate the metrics though - we should think about the input schema here more thoroughly at some point. As long as we keep the endpoint's name consistent we can change the schema freely imo

api/api/validators/evaluate_team_set_validator.py
```diff
@@ -0,0 +1,199 @@
...
+        # No teams should have the same id
+        team_ids = [team["id"] for team in team_set["teams"]]
+        if len(team_ids) != len(set(team_ids)):
+            raise ValueError("team_set[teams] contains teams with the same id")
+
+        # There should not be any duplicate students
+        student_ids = [
+            student["id"] for team in team_set["teams"] for student in team["students"]
+        ]
+        if len(student_ids) != len(set(student_ids)):
```
Opeyem1a, line 91: nit;

could use the `is_unique` util

api/api/views/evaluate_team_set.py
```diff
@@ -0,0 +1,94 @@
...
+        try:
+            request_data: Dict[str, Dict[str, Any]] = dict(request.data)
+
+            EvaluateTeamSetValidator(request_data).validate()
+
+            team_set_json = request_data["team_set"]
+            team_set = TeamSetSerializer().decode(team_set_json)
+            metrics = request_data["metrics"]
+
+            return_metrics: Dict[str, float] = {
```
Opeyem1a, line 41: it might be useful to return the team set as well as the metrics.

This way you don't need to match a specific request with a response to know what it was referring to. (Consider the async case where you send multiple teamsets for evaluation at the same time)

Just a thought though

api/api/validators/evaluate_team_set_validator.py
```diff
@@ -0,0 +1,199 @@
...
+
+        Schema([int]).validate(available_timeslots)
+        Schema(int).validate(timeslot_attribute_id)
+
+        if len(available_timeslots) != len(set(available_timeslots)):
+            raise ValueError("available_timeslots contains duplicate timeslots")
+
+        # all students' timeslots
+        all_students_timeslots = set()
+        for _teams in self.data["team_set"]["teams"]:
```
Opeyem1a, line 187: nit;

makes me sad when variables that are singular are named plurally :'(

api/api/validators/evaluate_team_set_validator.py
```diff
@@ -0,0 +1,199 @@
...
+            "available_timeslots"
+        ]
+        timeslot_attribute_id = common_time_availability_metric_params[
+            "timeslot_attribute_id"
+        ]
+
+        Schema([int]).validate(available_timeslots)
+        Schema(int).validate(timeslot_attribute_id)
+
+        if len(available_timeslots) != len(set(available_timeslots)):
```
Opeyem1a, line 182: nit;

`is_unique`

api/api/validators/evaluate_team_set_validator.py
```diff
@@ -0,0 +1,199 @@
...
+            # TODO: After deadline, make a get metric method like in AlgorithmRunner
+            # (https://github.com/Teamable-Analytics/algorithms/issues/369)
+            if metric == "avg_cosine_difference":
+                self._validate_avg_cosine_metric(params)
+            elif metric == "avg_cosine_similarity":
+                self._validate_avg_cosine_metric(params)
+            elif metric == "avg_solo_status":
+                self._validate_solo_status_metric(params)
+            elif metric == "common_time_availability":
+                pass
```
Opeyem1a, line 106: why not validate it?

api/api/validators/evaluate_team_set_validator.py
```diff
@@ -0,0 +1,199 @@
...
+            elif metric in ["project_coverage", "social_satisfaction"]:
```
Opeyem1a, line 107: why mention if not implemented?

api/api/views/evaluate_team_set.py
Opeyem1a: `ResponseWithMetadata` is a bit shaky of a paradigm for this reason, too. We should have the freedom to return multiple keys of data freely without the data label forcing us to return only one

Opeyem1a (changes requested):

api/api/views/evaluate_team_set.py
ketphan02: I can put a TODO

Opeyem1a (approved):

PR #371, opened 2024-02-25
Serializer should use id instead of _id

ketphan02:
- Fix `_id` -> `id`


Opeyem1a (changes requested):
Did you test how this works with the rest of the benchmarking though? We built serializers originally so that student data could be serialized and read again.

I don't really know if we need to delete the keys though, I don't see why they can't just be null in the API response.

ketphan02:
The `id` does work thanks to `json_dict.get("id", json_dict.get("_id"))`

Opeyem1a (approved):

PR #373, opened 2024-02-28
Refactor some API schema validation fields

ketphan02:


api/api/validators/evaluate_team_set_validator.py
```diff
@@ -12,7 +12,9 @@ class EvaluateTeamSetValidator(Validator):
     def validate(self):
         self._validate_schema()
         self.validate_team_set()
+        print(3)
```
Opeyem1a, line 15: 👀

api/api/validators/evaluate_team_set_validator.py
```diff
@@ -42,8 +43,8 @@ def validate_team_set(self):
             [
                 {
                     "id": int,
-                    Optional("name"): str,
-                    Optional("project_id"): int,
+                    Optional("name"): Or(str, None),
```
Opeyem1a, line 46: Q: is there not a way to do like Optional(str)? No worries if not just curious

Opeyem1a (changes requested):

Opeyem1a (approved):

api/ai/interfaces/algorithm_options.py
```diff
@@ -128,8 +128,8 @@ def get_schema() -> Schema:
                 SchemaOptional("enemy_behaviour"): Or(
                     *[behaviour.value for behaviour in RelationshipBehaviour]
                 ),
-                SchemaOptional("attributes_to_diversify"): List[int],
-                SchemaOptional("attributes_to_concentrate"): List[int],
+                SchemaOptional("attributes_to_diversify"): Or(list[int], []),
```
Opeyem1a, line 131: seems weird to me that the lower case is needed, and that we need to explicitly union set with it just to account for empty lists...

I'll trust you if you say so but seems odd

Opeyem1a (approved):

PR #382, opened 2024-02-28
COSC 341 Benchmarking Runs

Hedgemon4:
### Description
- These are the runs that we used to benchmark the 341 data for the paper.


benchmarking/runs/real_data/cosc_341/scenario1.py
```diff
@@ -0,0 +1,231 @@
...
+    DiversifyGenderMin2ConcentrateTimeSlotAndDiversifyYearLevel,
+)
+from benchmarking.runs.interfaces import Run
+from benchmarking.simulation.goal_to_priority import goals_to_priorities
+from benchmarking.simulation.insight import InsightOutput, Insight
+from benchmarking.simulation.simulation_set import SimulationSet
+from benchmarking.simulation.simulation_settings import SimulationSettings
+
+
+class Scenario1(Run):
```
Opeyem1a, line 37: nit;

it was a bad idea to name these runs "scenario", when we already have another type of object in this repo called Scenario, not back breaking, but it's icky

benchmarking/runs/real_data/cosc_341/scenario1.py
```diff
@@ -0,0 +1,231 @@
...
+                            os.path.join(
+                                os.path.dirname(__file__),
+                                "../../../..",
+                                "api/ai/external_algorithms/group_matcher_algorithm/group-matcher/run.py",
+                            )
+                        ),
+                    ),
+                ],
+            },
+        ).run(num_runs=1)
```
Opeyem1a, line 112: why is rhis one num_runs=1 and not =num_trials?

benchmarking/runs/real_data/cosc_341/scenario1.py
```diff
@@ -0,0 +1,231 @@
...
+                    insight_set, metrics[metric_name].name
+                )
+
+            for metric_name, average_metric in average_metrics.items():
+                if metric_name not in graph_data:
+                    graph_data[metric_name] = {}
+                for algorithm_name, value in average_metric.items():
+                    if algorithm_name not in graph_data[metric_name]:
+                        graph_data[metric_name][algorithm_name] = GraphData(
+                            x_data=[175],
```
Opeyem1a, line 171: at this point, 175 shows up so much that we should define it as a constant at the top of this function and give it a name (I'm assuming it's the class size)

Opeyem1a (changes requested):

benchmarking/runs/real_data/cosc_341/scenario1.py
Hedgemon4: I think I am going to leave it because it corresponds to something in the paper, but I can leave a comment to explain.

Hedgemon4: Or I can rename it if you prefer and leave a comment to explain what scenario it corresponds to.

benchmarking/runs/real_data/cosc_341/scenario1.py
Hedgemon4: This is because it was intended to only be run once on the real data to reproduce what an actual user would do

Hedgemon4: But I can still change it and just have it default to one instead

benchmarking/runs/real_data/cosc_341/scenario1.py
Hedgemon4: Yeah, it is. I can do that.

benchmarking/runs/real_data/cosc_341/scenario1.py
Hedgemon4: Oh wait this was 1 run because group matcher is deterministic, so the output never changes

Opeyem1a: Ah okay

Opeyem1a (approved):

PR #390, opened 2024-03-07
Add models and serializers for logging API calls

### LINKED ISSUE #389: Add models for logging API calls
Full feature described here: https://docs.google.com/document/d/1NblJCwjJxUUSGxgDBpmwfDNsTNS6V3YfAhRJio6uVd4/edit

justino599:
### Description

I am using model inheritance to account for future API endpoints. The base has the following attributes:
```
endpoint = models.CharField(max_length=100)
source = models.CharField(max_length=50, choices=APIUseCase.choices)
api_version = models.CharField(max_length=50)
data = models.JSONField()
```
Django uses a `jsonb` datatype in postgres for this which supports faster querying and indexing with a few downsides including not preserving key order.
For more reading:
- https://docs.djangoproject.com/en/5.0/ref/models/fields/#jsonfield
- https://pganalyze.com/blog/postgres-jsonb-django-python

For inserting to the json field:
```py
APILogobjects.create(
    endpoint="/generate",
    source=APISoource.INTERNAL_DEMO,
    api_version="1.0.0",
    generation_configuration={"limit": 100},  # Example data
)
APILog.objects.create(
    endpoint="/generate",
    source=APISource.NORMAL,
    api_version="1.0.0",
    generation_configuration={"limit": 1000, "old": True},  # Example data
)
```
For querying the json field:
```py
# Get all APILog objects with data of {"old": False} or not specified
APILog.objects.filter(
    data__old__isnull=True
) | APILog.objects.filter(data__old=False)
```

Closes #389


api/api/models/logs.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    Choices for the usecase field in the APILog model
+    """
+
+    NORMAL = "Normal", "Normal"
+    INTERNAL_DEMO = "Internal Demo", "Internal Demo"
+
+
+class APILog(BaseModel):
+    endpoint = models.TextField()
```
Opeyem1a, line 16: We can charfield this with max length of like 100 I think

api/api/models/logs.py
```diff
@@ -0,0 +1,23 @@
...
+    http_method = models.CharField(max_length=50)
```
Opeyem1a, line 17: This is actually a discrete set of options, no?
Also idk if we gain that much value from this field tbh, the kind of API request we want to log is almost all POST

api/api/models/logs.py
```diff
@@ -0,0 +1,23 @@
...
+    use_case = models.CharField(max_length=50, choices=APIUseCase.choices)
```
Opeyem1a, line 18: nit;

I would name this more like `source`

Opeyem1a (changes requested):

api/api/models/logs.py
justino599: We can have it default to POST, but I think it might be something that we might want in the future, but ofc we can't be sure of that. But defaulting it isn't really that helpful I guess since all of these records will be generated in one spot

api/api/models/logs.py
justino599: Ok, I'm good with that, I tried to come up with a better name, but just ended up settling with this

Opeyem1a (approved):

PR #393, opened 2024-03-09
Added Greedy Local Max Mutation

Hedgemon4:
### Description
- added a greedy version of local max which randomly takes two teams, places the students in a list, and then places each student on to the team that increases the score the most


Opeyem1a (approved):
Mostly a test nit

PR #395, opened 2024-03-11
remove unused api logs generate serializer

Opeyem1a:


justino599 (approved):
Good catch, I totally forgot to remove that

PR #406, opened 2024-03-16
Refactor Mutations

Hedgemon4:
### Description
- Refactored mutations to extend a class instead of each of them being defined seperately.
- Encapsulated mutations to call and number of times to call in mutation set instead of priority algorithm class.


api/ai/priority_algorithm/mutations/mutation_set.py
```diff
@@ -0,0 +1,41 @@
+from typing import List, Dict, Tuple
+
+from api.ai.priority_algorithm.custom_dataclasses import PriorityTeamSet
+from api.ai.priority_algorithm.mutations.mutation import Mutation
+from api.ai.priority_algorithm.priority.interfaces import Priority
+from api.dataclasses.student import Student
+
+
+class MutationSet:
```
Opeyem1a, line 9: Quick review notes:
https://www.loom.com/share/7488f57b699e4c48bb68916b1d6eb2a5

Opeyem1a (changes requested):

api/ai/priority_algorithm/mutations/interfaces.py
```diff
@@ -0,0 +1,69 @@
...
+        self.mutations = mutations
+
+    def generate_mutations(
+        self,
+        team_set: PriorityTeamSet,
+        priorities: List[Priority],
+        student_dict: Dict[int, Student],
+    ) -> List[PriorityTeamSet]:
+        mutated_team_sets: List[PriorityTeamSet] = []
+        for mutation in self.mutations:
```
Opeyem1a, line 58: nit;

couldn't this just be list comprehension?

api/ai/priority_algorithm/mutations/interfaces.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def mutate_one(
+        self,
+        priority_team_set: PriorityTeamSet,
+        priorities: List[Priority],
+        student_dict: Dict[int, Student],
+    ) -> PriorityTeamSet:
+        raise NotImplementedError
+
+    def num_mutations(self) -> int:
```
Opeyem1a, line 36: pls make a property

Opeyem1a: actually wait why do we need this to be a function?

api/ai/priority_algorithm/mutations/local_max_double_random.py
```diff
@@ -0,0 +1,36 @@
...
+        combinations of students on those teams. It replaces one of the teams with the team with the highest score found,
+        and puts all other students on the other team. This implementation does not account for if the second team's score
+        increases or decreases.
+        """
+        available_priority_teams = get_available_priority_teams(priority_team_set)
+        try:
+            if len(available_priority_teams) < 2:
+                return priority_team_set
+            team_1, team_2 = random.sample(available_priority_teams, 2)
+            local_max(team_1, team_2, priorities, student_dict)
```
Opeyem1a, line 32: we should iron out what is and isn't a pure function at some point, since a lof these do directly modify the team set passed in

api/ai/priority_algorithm/priority_algorithm.py
```diff
@@ -78,7 +78,9 @@ def generate(self, students: List[Student]) -> TeamSet:
         ):
             new_team_sets: List[PriorityTeamSet] = []
             for team_set in team_sets:
-                new_team_sets += self.mutate(team_set)
+                new_team_sets += self.algorithm_config.MUTATIONS.generate_mutations(
```
Opeyem1a, line 81: Nothing wrong, but might still be useful to leave the self.mutate() function intact for now.

Like just call self.mutate() and put this line of code in there. Idk why but I have a feeling we should keep this unneeded layer of abstraction in case of future work

api/ai/priority_algorithm/mutations/interfaces.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class MutationSet:
+    def __init__(self, mutations: List[Mutation]):
+        """
+        A set of mutations to be run in a single iteration of the priority algorithm.
+
+        Parameters
+        ----------
+        mutations - A list of tuples of a Mutation to the number of times it should be called.
```
Opeyem1a, line 47: no longer the real type

benchmarking/runs/concentrate_gpa.py
```diff
@@ -92,7 +94,9 @@ def start(self, num_trials: int = 10, generate_graphs: bool = True):
                         PriorityAlgorithmConfig(),
                         PriorityAlgorithmConfig(
                             name="local_max",
-                            MUTATIONS=[(mutate_local_max, 1), (mutate_random_swap, 2)],
+                            MUTATIONS=MutationSet(
+                                [(LocalMaxMutation(), 1), (RandomSwapMutation(), 2)]
```
Opeyem1a, line 98: Wait wouldn't it be more like `[LocalMaxMutation(1), RandomSwapMutation(2)]` now?

Opeyem1a (changes requested):

api/ai/priority_algorithm/priority_algorithm.py
```diff
@@ -130,6 +111,18 @@ def _unpack_priority_team_set(self, priority_team_set: PriorityTeamSet) -> TeamS
...
         return TeamSet(teams=teams)
 
+    def mutate(
+        self,
+        team_set: PriorityTeamSet,
+    ) -> List[PriorityTeamSet]:
+        return [
+            mutated_set
+            for mutation in self.algorithm_config.MUTATIONS
+            for mutated_set in mutation.mutate(
```
Opeyem1a, line 121: the ordering of the 2 `for` parts seems backwards for this list comprehension to work, maybe a quick sanity test that it always returns a list of PriorityTeamSet and not a list of lists or something?

Opeyem1a (approved):
One semi-nit, approved in case I'm wrong about it but wise to double check

api/ai/priority_algorithm/priority_algorithm.py
justino599: I added a test for it :)

PR #408, opened 2024-03-19
Log the /generate/teams endpoint

### LINKED ISSUE #399: Add a logger to create log entries in the database on API calls
More details at https://docs.google.com/document/d/1NblJCwjJxUUSGxgDBpmwfDNsTNS6V3YfAhRJio6uVd4/edit#heading=h.qjpa9i1e5zyw

justino599:
- Adds an API version constant
- Logs calls to the /generate/teams endpoint
- Adds a query param to endpoint, `sandbox=[boolean]`, that, if true, will cause the call to not be logged
- Adds a list of domains considered to be from an internal demo

Closes #399


api/api/views/generate_teams.py
```diff
@@ -1,24 +1,48 @@
...
+                        "algorithm_options": request_data.get("algorithm_options"),
+                        "team_generation_options": request_data.get(
+                            "team_generation_options"
+                        ),
+                    },
+                    source=(
+                        APISource.INTERNAL_DEMO
+                        if request.META.get("REMOTE_ADDR") in api.INTERNAL_DEMO_DOMAINS
+                        else APISource.NORMAL
+                    ),
```
justino599, line 43: I'm not sure if I want this logic to be here or in the logger, let me know if you have thoughts on this.

api/api/utils/logger.py
```diff
@@ -0,0 +1,17 @@
+from typing import List, Dict, Union
+
+import api
+from api.api.models import APISource, APILog
+
+
+class APILogger:
+    @staticmethod
+    def log(
```
Opeyem1a, line 9: we kinda don't need this but I see the vision.

tldr this is basically a "repository service" pattern, but Django models already are basically repository patterns.

basically, if you want this you can just add an extra method to the `APILog` model itself

api/api/views/generate_teams.py
```diff
@@ -1,24 +1,48 @@
...
+    def generate_teams(self, request: Request):
         try:
             request_data = dict(request.data)
 
             # validate data schema
             GenerateTeamsValidator(request_data).validate()
             input_data = GenerateTeamsDataLoader(request_data).load()
 
+            # Lowercase the params
+            params = {k.lower(): v for k, v in request.query_params.items()}
```
Opeyem1a, line 27: are they not already lowercase?

api/api/views/generate_teams.py
Opeyem1a: no, this makes sense

api/api/views/generate_teams.py
```diff
@@ -1,24 +1,48 @@
...
+                    endpoint="/generate/teams",
+                    data={
+                        "algorithm_options": request_data.get("algorithm_options"),
+                        "team_generation_options": request_data.get(
+                            "team_generation_options"
+                        ),
+                    },
+                    source=(
+                        APISource.INTERNAL_DEMO
+                        if request.META.get("REMOTE_ADDR") in api.INTERNAL_DEMO_DOMAINS
```
Opeyem1a, line 41: assuming you've tested that request.META.REMOTE_ADDR is the string you want it to be

Opeyem1a (changes requested):

api/api/views/generate_teams.py
justino599: Locally it returned 127.0.0.1 when I tried it, so it would seem that it does

api/api/views/generate_teams.py
justino599: No, if you entered "?Sandbox=true" it comes through as "Sandbox", not "sandbox"

api/api/utils/logger.py
justino599: ok, I've moved the function to the APILog model and checked that it still works

Opeyem1a (approved):

PR #410, opened 2024-03-25
Pass min/max team size to mutations

### LINKED ISSUE #409: Implement the ability to have variable sized teams
Mutations should be given the min and max team sizes. Teams should be treated as always being at the max team size where you have the real students and then however many fake, or "null", students are needed to bring the team to full. Then, the mutation should be able to swap "null" students as if they were real students, but verifying that there area actually enough real students are still in the team to meet the min team size requirement.

This PR does not address the problem of changing the total number of teams.

justino599:
Closes #409


api/ai/priority_algorithm/mutations/interfaces.py
```diff
@@ -15,12 +15,16 @@ def mutate(
         priority_team_set: PriorityTeamSet,
         priorities: List[Priority],
         student_dict: Dict[int, Student],
+        min_team_size: int,
```
Opeyem1a, line 18: why not pass the whole team gen options object instead?

Opeyem1a (changes requested):

benchmarking/simulation/mock_algorithm.py
```diff
@@ -35,13 +35,20 @@
...
+        initial_teams: List[TeamShell] = None,
+        min_team_size: int = None,
+        max_team_size: int = None,
     ) -> TeamGenerationOptions:
         _num_teams = len(initial_teams) if initial_teams else num_teams
-        min_team_size = num_students // _num_teams
-        max_team_size = (
-            min_team_size if num_students % _num_teams == 0 else min_team_size + 1
-        )
+        if min_team_size and not max_team_size or max_team_size and not min_team_size:
```
Opeyem1a, line 45: I know AND is done first over OR, but maybe some brackets here since it's not immediately clear imo

benchmarking/simulation/simulation_settings.py
```diff
@@ -10,19 +10,44 @@ class SimulationSettings:
...
+                raise ValueError("Min team size must be smaller than max team size")
+            num_students = self.student_provider.num_students
+            num_teams_local = self.num_teams or len(self.initial_teams_provider.get())
+            min_students = num_teams_local * self.min_team_size
+            if min_students > num_students:
+                raise ValueError(
+                    "The min team size is too big to fill all teams with the given students"
+                )
+            max_students = num_teams_local * self.max_team_size
+            if max_students < num_students:
```
Opeyem1a, line 50: Good thinking with this

Opeyem1a (approved):
Nits

PR #412, opened 2024-03-27
Refactor Greedy Local Max to Work With N Teams

Hedgemon4:
### Description
- refactor greedy local max to accept a parameter N to choose the number of teams to mutate instead of just two


api/ai/priority_algorithm/mutations/greedy_local_max.py
```diff
@@ -13,6 +13,14 @@
 
 
 class GreedyLocalMaxMutation(Mutation):
+    def __init__(self, num_mutations: int = 1, number_of_teams: int = 2):
+        super().__init__(num_mutations)
+        self.number_of_teams = number_of_teams
+
+    def validate(self):
```
Opeyem1a, line 20: when do we call this method btw?

api/ai/priority_algorithm/mutations/greedy_local_max.py
```diff
@@ -13,6 +13,14 @@
 
 
 class GreedyLocalMaxMutation(Mutation):
+    def __init__(self, num_mutations: int = 1, number_of_teams: int = 2):
+        super().__init__(num_mutations)
+        self.number_of_teams = number_of_teams
```
Opeyem1a, line 18: nit;

unsure how it can be better, but this variable name is kinda hard to interpret - perhaps drop a comment somewhere about it, but wasn't too hard to understand from usage

api/ai/priority_algorithm/mutations/greedy_local_max.py
```diff
@@ -31,37 +39,41 @@ def mutate_one(
             priority_team_set
         )
         try:
```
Opeyem1a, line 41: github won't let me add a comment to it directly, but if you're going to write a comment for a method, you gotta update that comment when you change the method

Opeyem1a (changes requested):
Minor changes

Opeyem1a (approved):
lgtm

PR #415, opened 2024-03-28
Add team size mutation that minimizes team disruption

### LINKED ISSUE #413: Add another team size mutation minimizing team disruption

justino599:
Closes #413

### DIFF
--- a/api/ai/priority_algorithm/mutations/team_size_low_disruption.py
+++ b/api/ai/priority_algorithm/mutations/team_size_low_disruption.py
@@ -1,0 +1,95 @@
+from typing import List, Dict, Tuple
+
+import numpy as np
+
+from api.ai.interfaces.team_generation_options import TeamGenerationOptions
+from api.ai.priority_algorithm.custom_dataclasses import PriorityTeamSet, PriorityTeam
+from api.ai.priority_algorithm.mutations.interfaces import Mutation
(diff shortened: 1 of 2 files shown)


api/ai/priority_algorithm/mutations/team_size_low_disruption.py
```diff
@@ -0,0 +1,108 @@
...
+from api.ai.priority_algorithm.mutations.utils import get_available_priority_teams
+from api.ai.priority_algorithm.priority.interfaces import Priority
+from api.dataclasses.student import Student
+
+
+class TeamSizeLowDisruptionMutation(Mutation):
+    """
+    Mutation to adjust team sizes within the bounds.
+
+    The idea is to first randomly generate the sizes of the new teams (with approx uniform distribution).
```
justino599, line 17: Should we generate new team sizes with a uniform dist, or should it be a normal dist? I kinda thought a uniform would be more interesting so that the edge sizes (closer to min/max size) are just as common as those in the middle. But maybe we want to try sticking to the middle more ¯\_(ツ)_/¯

Opeyem1a: idk, you can try the other version as an experiment, the uniform logic makes sense to me though

Opeyem1a (approved):

PR #417, opened 2024-03-28
Add csv to json converter

### LINKED ISSUE #416: CSV to JSON converter for Bowen
CSV with columns
- name
- atribute1, attribute2, etc...
- friends (list of student names)
- enemies

Should convert to the JSON students that the API call takes.

justino599:
### Description
Converts a CSV list of students with attributes to a JSON object compatible with the algorithms API.

Format:
- There are three special column names, 'student_name', 'friends', and 'enemies'. These must be spelled exactly as provided to be recognized. student_names is a required column.
- Any other column is considered an attribute and will be numbered in the order of appearance in the output json.
- Values for an attribute are kept as a number if it is provided as a number in the csv, otherwise a number is assigned to each unique non-integer value.

### Example

This csv:
```csv
student_name,attr1,attr2,these name's don’t matter,friends,enemies
s1,2,0,red,"s2,s3",
s2,2,3,blue,s1,s3
s3,23,3,"red,blue",s1,
s4,,,,,
```

produces this json:
```json
[
    {
        "id": 0,
        "name": "s1",
        "attributes": {
            "0": [
                2
            ],
            "1": [
                0
            ],
            "2": [
                0
            ]
        },
        "relationships": {
            "1": "friend",
            "2": "friend"
        },
        "project_preferences": []
    },
    {
        "id": 1,
        "name": "s2",
        "attributes": {
            "0": [
                2
            ],
            "1": [
                3
            ],
            "2": [
                1
            ]
        },
        "relationships": {
            "0": "friend",
            "2": "enemy"
        },
        "project_preferences": []
    },
    {
        "id": 2,
        "name": "s3",
        "attributes": {
            "0": [
                23
            ],
            "1": [
                3
            ],
            "2": [
                0,
                1
            ]
        },
        "relationships": {
            "0": "friend"
        },
        "project_preferences": []
    },
    {
        "id": 3,
        "name": "s4",
        "attributes": {},
        "relationships": {},
        "project_preferences": []
    }
]
```

Closes #416


api/utils/studentCSVToJson.py
```diff
@@ -0,0 +1,130 @@
...
+                    student["relationships"][related_student] = "friend"
+                elif (
+                    student["relationships"][related_student]
+                    == Relationship.ENEMY.value
+                ):
+                    student["relationships"][related_student] = "enemy"
+                else:
+                    student["relationships"][related_student] = "default"
+
+        return json.dumps(students_dict, indent=4)
```
justino599, line 130: I think this piece of code needs an explanation.

K's API requires that student relationships be in the format `relationships: { "sid" : ["friend" | "enemy" | "default"] }`. The `StudentSerializer` obviously doesn't output this format since Students map sids to Relationship objects.

This code first converts student objects to JSON, then converts to a python dict. Then, I change the Relationship value to the string the the API wants, then since it's a dict and not a Student list, I can convert back to JSON without using the StudentSerializer

api/utils/studentCSVToJson.py
Opeyem1a, line 1: name this file in snake case

Opeyem1a (approved):
Tbh high level I only kind of get why this is here and/or needed, I'm assuming some use case came up. I have no real comments since it doesn't really interact with this repo at all. I'm almost tempted t say it's the kind of script that doesn't need to be merged at all

Opeyem1a (changes requested):
oops shouldn't have approved, the snake case thing is important

justino599:
> Tbh high level I only kind of get why this is here and/or needed, I'm assuming some use case came up. I have no real comments since it doesn't really interact with this repo at all. I'm almost tempted t say it's the kind of script that doesn't need to be merged at all

Yeah we can leave it unmerged if you like I'm not totally sure what Bowen's plans were with this, I think it's for some sort of demo

Opeyem1a (approved):

Opeyem1a:
It's here now so probably mergeable, it is what it is

PR #426, opened 2024-04-12
Production deployment fix

ketphan02:


entrypoint.sh
```diff
@@ -0,0 +1,21 @@
...
+echo "Flush the manage.py command it any"
+
+while ! python manage.py flush --settings=api.api.settings.prod --no-input 2>&1; do
+  echo "Flusing django manage command"
+  sleep 3
+done
+
+echo "Migrate the Database at startup of project"
+
+echo api/api/settings/prod.py
```
ketphan02, line 11:
```suggestion

echo "Migrate the Database at startup of project"
```

utils/api_auth.py
```diff
@@ -24,9 +24,10 @@ def check_hash(api_token, encrypted_api_token) -> bool:
     """
     Checks if the API token matches the encrypted API token.
     """
+    print(api_token, encrypted_api_token)
```
Opeyem1a, line 27: nit;

Opeyem1a (changes requested):

api/ai/interfaces/algorithm_options.py
```diff
@@ -128,8 +128,14 @@ def get_schema() -> Schema:
...
-                SchemaOptional("attributes_to_diversify"): Or(List[int], []),
-                SchemaOptional("attributes_to_concentrate"): Or(List[int], []),
+                SchemaOptional("attributes_to_diversify"): And(
+                    list,
+                    lambda _data: all(isinstance(i, int) for i in _data),
+                ),
+                SchemaOptional("attributes_to_concentrate"): And(
+                    list,
+                    lambda _data: all(isinstance(i, int) for i in _data),
+                ),
```
keyvankhademi, line 138: I'm curious
1. What was the issue with the old types?
2. How is it related to production deployment?

api/api/settings/prod.py
```diff
@@ -29,7 +29,8 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = False
 
-ALLOWED_HOSTS = list(os.environ.get("DJANGO_ALLOWED_HOSTS", "").split(" "))
```
keyvankhademi, line 32: I don't think you should remove ALLOWED_HOSTS!

api/api/settings/prod.py
```diff
@@ -29,7 +29,8 @@
...
+CORS_ALLOW_HEADERS = "*"
```
keyvankhademi, line 32: allow all headers could potentially cause security issues in prod. Please specify exactly the headers you need.

api/api/settings/prod.py
```diff
@@ -76,6 +77,18 @@
...
+    "default": {
+        "ENGINE": os.environ.get("DB_ENGINE", "django.db.backends.postgresql"),
+        "NAME": os.environ.get("DB_DATABASE", os.environ.get("RDS_DB_NAME")),
+        "USER": os.environ.get("DB_USER", os.environ.get("RDS_USERNAME")),
+        "PASSWORD": os.environ.get("DB_PASSWORD", os.environ.get("RDS_PASSWORD")),
+        "HOST": os.environ.get("DB_HOST", os.environ.get("RDS_HOSTNAME")),
+        "PORT": os.environ.get("DB_PORT", os.environ.get("RDS_PORT")),
+    }
+}
+
```
keyvankhademi, line 91: I thought the algorithms api was stateless.
@Opeyem1a am I missing something?

utils/api_auth.py
```diff
@@ -4,7 +4,7 @@
 from pathlib import Path
 import dotenv
 
-env_file = next((Path(__file__).parent.parent / "env").glob("*.env"))
+env_file = next((Path(__file__).parent.parent / "env").glob(".env*"))
```
keyvankhademi, line 7: doesn't this cause issues if there are multiple env files present in the folder? It seems like it will randomly choose 1.

utils/api_auth.py
```diff
@@ -36,5 +37,6 @@ def is_api_key_valid(api_key) -> bool:
     """
     allowed_api_keys = os.getenv("API_AUTH_ALLOWED_KEYS").split(" ")
     return any(
-        check_hash(api_key, allowed_api_key) for allowed_api_key in allowed_api_keys
+        check_hash(api_key, bytes.fromhex(allowed_api_key))
+        for allowed_api_key in allowed_api_keys
```
keyvankhademi, line 41: @Opeyem1a
We are only calling the apis from our backend, right?

api/api/settings/prod.py
Opeyem1a: @keyvankhademi 

Intent is to store analytics data on calls to the algorithm API

utils/api_auth.py
Opeyem1a: @keyvankhademi 

Currently the list of consumers is:
- Our backend
- Our simple API frontend (something for demos)

We intend this API to be exgtensible for consumption at other institutions (i.e. Harvard Business School discussed wanting to integrate with our algorithms and have already developed an internal tool for this that could connect to our API instead of whatever they are currently doing)

api/ai/interfaces/algorithm_options.py
ketphan02: This was supposed to be a separate PR but I put it here to conveniently deploy it. It has been a while, but I think the reason for this was this throws an error with `List` because it does not recognize it and wants to use `list` instead.

api/api/settings/prod.py
ketphan02: Currently storing request metadata (IP, time, URL path)

PR #428, opened 2024-04-23
Run to compare AI algorithms

ketphan02:
This run compare priority with DRR, MRR, GEG, and GROUP_MATCHER


benchmarking/evaluations/metrics/num_teams.py
```diff
@@ -0,0 +1,7 @@
+from api.dataclasses.team_set import TeamSet
+from benchmarking.evaluations.interfaces import TeamSetMetric
+
+
+class NumberOfTeams(TeamSetMetric):
```
ketphan02, line 5: Note: this is not really a metric, but really handy to check the number of teams in 1 team set, especially when algorithms do not return the same number of teams

PR #436, opened 2024-09-20
Weight Run Modifications

colton456p:
### The new code can be found in **manual_run** and the changes include:
- Rewrote the data provider script.
- Rewrote the weight run script and changed the name to manual_priority_run.py.
- Added proper documentation to the code for feauture use as this was originally missing.
- Added Attributes class.
- Added map_columns script which sets  any string rows to their corresponding integer value
- Added variables script in order to keep all user inputted info about the csv data in the same place.

### Main files modified/created:
- attributes.py
- data_provider.py
- default_scenario.py
- manual_priority_run.py
- map_columns.py
- variables.py


benchmarking/runs/csv_weight_run_revised/attributes.py
Opeyem1a, line 1: This file is my least favourite, recorded a video about it - https://www.loom.com/share/2fbd4127026c4c9d85b03b1105ec2594

benchmarking/runs/csv_weight_run_revised/weight_run_revised.py
```diff
@@ -0,0 +1,159 @@
...
+
+        output_file = os.path.join(output_dir, "result.csv")
+
+        with open(output_file, "w+", newline="") as f:
+            writer = csv.writer(f)
+            for row in data_fields:
+                writer.writerow(row)
+
+
+class DefaultScenario(Scenario):
```
Opeyem1a, line 124: I would have this live in a different file

benchmarking/runs/csv_weight_run_revised/weight_run_revised.py
```diff
@@ -0,0 +1,159 @@
...
+        This function is meant to create a list of scenerio goals to be followed for team formation.
+
+        Returns:
+            List[Goal]: A list of goals for team formation.
+        """
+        return [
+            # Concentrate on student timeslot availability
+            DiversityGoal(
+                DiversifyType.CONCENTRATE,
+                ScenarioAttribute.TIMESLOT_AVAILABILITY.value,
```
Opeyem1a, line 141: Applies to every other file too, but it's dangerous to be using ScenarioAttribute at all in any of this

benchmarking/runs/csv_weight_run_revised/weight_run_revised.py
```diff
@@ -0,0 +1,159 @@
...
+        team_set = list(artifact.values())[0][0][0]
+
+        insight_output_set = Insight.get_output_set(artifact, metrics)
+        print(insight_output_set)
+
+        # Enter the data fields from the provided CSV file
+        data_fields = [
+            ["ResponseId", "Q8", "Q4", "Q5", "zPos", "TeamSizeViolation", "TeamId"]
+        ]
+        for team in team_set.teams:
```
Opeyem1a, line 68: I think all code related to creating the output csv should be in it's own function

benchmarking/runs/csv_weight_run_revised/weight_run_revised.py
```diff
@@ -0,0 +1,159 @@
...
+                name="Score Cosine Difference",
+                attribute_filter=[Attributes.SCORE.value],
+            ),
+            AverageCosineDifference(
+                name="Timeslot Cosine Difference",
+                attribute_filter=[ScenarioAttribute.TIMESLOT_AVAILABILITY.value],
+            ),
+        ]
+        student_provider = DataProvider()
+        print(student_provider)
```
Opeyem1a, line 38: All of these print statements for debugging need to be removed before we merge this

benchmarking/runs/csv_weight_run_revised/weight_run_revised.py
```diff
@@ -0,0 +1,159 @@
...
+        team_size = 4  # enter the desired team size
```
Opeyem1a, line 39: This is another random constant that the user would have to open this file for. I would recommend finding all of these kinds of constants and making it obvious where they all live

benchmarking/runs/csv_weight_run_revised/weight_run_revised.py
```diff
@@ -0,0 +1,159 @@
...
+        artifact: SimulationSetArtifact = SimulationSet(
+            settings=SimulationSettings(
+                num_teams=ceil(student_provider.num_students / team_size),
+                scenario=scenario,
+                student_provider=student_provider,
+                # cache_key=f"weight_run_for_bowen/weight_run_2/",
+                cache_key=f"csv_weight_run_revised/weight_run_revised/",
```
Opeyem1a, line 46: it doesn't need a cache key at all, providing one saves it to the cache, leaving it undefined means the output is not saved.

In this instance, we don't need to cache (which means saving a .json of the output to your computer), because we're not planning to re-run it 100s of times for testing.

benchmarking/runs/csv_weight_run_revised/weight_run_revised.py
```diff
@@ -0,0 +1,159 @@
...
+                AlgorithmType.PRIORITY: [
+                    PriorityAlgorithmConfig(
+                        MAX_KEEP=15,  # max num of solutions the algorithm is allowed to keep
+                        MAX_SPREAD=30,  # max num of team sets genereated by mutation at each step
+                        MAX_ITERATE=30,  # max num of iterations the algorithm is allowed to run
+                        MAX_TIME=100000,  # max time in seconds the algorithm can run for.
+                    ),
+                ]
+            },
+        ).run(num_runs=num_trials)
```
Opeyem1a, line 58:
```suggestion
        ).run(num_runs=1)
```

It'll never be more than 1 generation when you're doing a manual generation

Opeyem1a (changes requested):
Some guiding comments for you. Feel free to ask any questions

manual_run/weight_run.py
```diff
@@ -0,0 +1,127 @@
...
+            ),
+        ]
+        student_provider = DataProvider()
+        artifact: SimulationSetArtifact = SimulationSet(
+            settings=SimulationSettings(
+                num_teams=ceil(student_provider.num_students / Variables.team_size),
+                scenario=scenario,
+                student_provider=student_provider,
+                cache_key=f"csv_weight_run/weight_run/",
+            ),
```
colton456p, line 46: If I remove this then the results don't print.

Opeyem1a: What do you mean? Like if you remove the cache key? Or the entire SimulationSet creation?

colton456p: Sorry ignore this. I had an error which is now fixed I didn't mean to publish that comment

manual_run/map_columns.py
```diff
@@ -0,0 +1,53 @@
...
+                if row[i] in Variables.attribute_options[head]:
+                    int_value = Variables.attribute_options[head].index(row[i])
+                attribute_options_int[head] = int_value
+            elif head in Variables.attribute_handlers:
+                int_value = Variables.attribute_handlers[head](row[i])
+                attribute_options_int[head] = int_value
+        return attribute_options_int
+
+    @staticmethod
+    def revert_timeslot(time_slot: int) -> str:
```
Opeyem1a, line 32: Why do we need these revert methods again?

manual_run/variables.py
```diff
@@ -0,0 +1,36 @@
...
+            "I am interested in being a peer tutor in BIOC 202",
+        ],
+        "Q5": ["1", "2 to 3", "3+"],
+    }
+
+    attribute_handlers: Dict[str, Callable] = {
+        "z-score": lambda x: 1 if float(x) >= 0 else 0
+    }
+
+    # Enter the name of the columns from the provided CSV file
```
Opeyem1a, line 29: Like all the columns in the provided CSV file? If it's all of them then why do we need this?

manual_run/weight_run.py
Opeyem1a, line 1: We named this file weight_run, but the algorithm it runs is the Priority algorithm

simulation_cache/csv_weight_run/weight_run/AlgorithmType.PRIORITY-default.json
Opeyem1a, line 1: delete these JSON files, they bloat the PR and are not needed

manual_run/weight_run.py
```diff
@@ -0,0 +1,118 @@
...
+                    tutor_preference
+                )
+                data_fields_input_3 = MapColumns.revert_group_size(group_size)
+
+                # Process score and team size violation
+                positive_z = "1" if attributes[Attributes.SCORE.value][0] == 1 else "0"
+                team_size = len(team.students)
+
+                # Can be customize to fit the team size preference
+                team_size_violation = (
```
Opeyem1a, line 82: you're right I dislike this, kinda care a lot less above this create_csv function and can learn to accept it if we're just choosing to do something greedy in the returned CSV. We should ponder how needed it is though

manual_run/README.md
Opeyem1a, line 1: I skimmed, but I feel like this file is incorrect now

manual_run/input_csv/README.md
Opeyem1a, line 1: Do we need this file or this directory? Can't we just have `manual_run/input.csv`? Not sure what we gain from the added folder

manual_run/map_columns.py
```diff
@@ -0,0 +1,53 @@
+from typing import Dict, List
+
+from manual_run.variables import Variables
+
+
+class MapColumns:
```
Opeyem1a, line 6: The pythonic way of doing this is to realize that any time you have a class where the class's only purpose is to provide static methods, you probably don't need a class and could have just defined the functions in this file without the class

Opeyem1a (changes requested):

manual_run/data_provider.py
```diff
@@ -0,0 +1,73 @@
+import csv
+from enum import Enum
+from os import path
+from typing import Dict
+
+import numpy as np
+
+# from api.models.student import Student
```
Opeyem1a, line 8: dead comment

manual_run/data_provider.py
```diff
@@ -0,0 +1,73 @@
...
+
+class DataProvider(StudentProvider):
+    def __init__(self):
+        self._sid_map: Dict[int, str] = {}
+        self._num_students = Variables.num_students
+        self._max_project_preferences_per_student = 0
+
+    @property
+    def num_students(self) -> int:
+        return self._num_students
```
Opeyem1a, line 24:
```suggestion
        return Variables.num_students
```

manual_run/data_provider.py
```diff
@@ -0,0 +1,73 @@
...
+from benchmarking.data.interfaces import StudentProvider
+from manual_run.attributes import Attributes
+from manual_run.map_columns import MapColumns
+from manual_run.variables import Variables
+
+
+class DataProvider(StudentProvider):
+    def __init__(self):
+        self._sid_map: Dict[int, str] = {}
+        self._num_students = Variables.num_students
```
Opeyem1a, line 19:
```suggestion
```

nitpick;

No need to set self._num_students if it's only used in 1 place. See the code suggestion below

manual_run/variables.py
colton456p: These were simply the important data fields to return that relate to the group formation. In reality the doc provided had the following columns (most of which dont need to be returned to the prof to identify the newly formed groups): ResponseId, Q8, Lecture, Midterm 1, Assignment 1, Assignment 2, Assignment 3, Assignment 4, Assignment 5, Assignment 6, Current points, Final points, etc...

manual_run/weight_run.py
colton456p: I don't personally see much of a use case for this **team_size_violation** attribute as it is very specific to each situation. I assuming that you're kinda thinking the same thing here. Would be this something I could remove?

manual_run/map_columns.py
colton456p: The only way I can think of reverting these integer values back to their string value would be by calling a function to do so. I over looked how simple this would be to break down into one function. Is doing this in a function like the following ok?
```
def revert_value(attribute_key: str, value: int) -> str:
    """Revert integer value to the original string based on the attribute key."""
    if attribute_key in Variables.attribute_options:
        options = Variables.attribute_options[attribute_key]
        if value < len(options):
            return options[value]
    return ""
```

manual_run/variables.py
Opeyem1a: So the goal is like "only deal with this column if it's named in this array"? If so, where is the logic that "filters" out the irrelevant columns?

manual_run/weight_run.py
Opeyem1a: I get the vibe Bowen probably wants it still so I'd just leave it in, maybe toss a comment so someone in the future knows they can remove it without breaking anything

manual_run/input.csv
Opeyem1a, line 1: Remove from PR

manual_run/result.csv
Opeyem1a, line 1: Remove from PR

manual_run/data_provider.py
```diff
@@ -0,0 +1,63 @@
...
+    def __init__(self):
+        self._max_project_preferences_per_student = 0
+
+    @property
+    def num_students(self) -> int:
+        return Variables.num_students
+
+    @property
+    def max_project_preferences_per_student(self) -> int:
+        return self._max_project_preferences_per_student
```
Opeyem1a, line 25: I forgot to mention but this is also just always going to be 0 for now

colton456p: I believe we talked about this during our meeting and I think you may have just missed it because just above I have it set to always be zero.

image

Opeyem1a: Ye I mean like there's no reason to have another variable just to set it to 0. Nothing is wrong, it's just simpler to have the method return 0 instead of return a variable that always equals 0

manual_run/variables.py
colton456p: Given the provided attributes in the two dictionaries **attribute_options** and **attribute_handlers** then when a row is processed it will only process the data in those columns. Given that that information is already provided I have now redone this to automatically create those rows based on the data in the two dictionaries. So now the data_fields list is not present in the variables.py script. Here's how it looks now in the **create_csv** function inside of the manual_priority_run.py script:

```
data_fields = [["ResponseId"]]
        for key in Variables.attribute_options:
            data_fields[0].append(key)
        for key in Variables.attribute_handlers:
            data_fields[0].append(key)
        data_fields[0].append("TeamSizeViolation")
        data_fields[0].append("TeamId")
```

Opeyem1a (approved):

Opeyem1a:
Remember to "squash and merge" when you merge this into main

